PR #28, opened 2018-08-10
Move templates into the root

justinvp:
Restructures the repo to get rid of the unnecessary templates sub directory and moves the build and scripts directories under _tools.


justinvp:
Going to merge this as the upcoming CLI changes depend on it. I can address any feedback subsequently.

justinvp:
@lukehoban FYI

PR #30, opened 2018-08-15
Update templates for 0.15.0

justinvp:
- Update dependencies to 0.15.0.
- Upgrade TypeScript templates to use the new first-class support.
- Disable "publish to S3" script. 0.15.0 pulls templates from GH now. Older versions of the CLI will see the last published templates on S3.
- Remove unneeded `.pulumi.template.yaml` files. They are no longer needed for 0.15.0 (which ignores them).


.travis.yml
```diff
@@ -1,4 +1,4 @@
-if: branch = master OR tag IS present
+if: branch = publish
```
justinvp: Only run the Travis build (which just runs the publish to S3 script) from the `publish` branch. This branch doesn't exist yet. If we ever need to update a template for an older CLI, we can create the branch and use it to publish changes to legacy templates.

ellismg:
I wonder if we should keep the tsconfig files for now?  Then being present do provide a nicer VS Code experience.

justinvp:
> I wonder if we should keep the tsconfig files for now? Then being present do provide a nicer VS Code experience.

Are there any changes to them that we should make, or are they good as-is? (e.g. just wondering about things like `"outDir": "bin"`).

PR #31, opened 2018-08-15
Openstack Typescript template

Frassle:


openstack-typescript/.gitignore
```diff
@@ -0,0 +1,3 @@
+/bin/
+/node_modules/
+/.pulumi/
```
ellismg: Mind deleting this line?  I'm removing the other copies in #32, it is no longer needed.

Frassle: Sure no problem.

PR #33, opened 2018-08-15
Add minimal Kubernetes template

justinvp:
I removed the `// Parameters` comments, kept `name` and `image` `const`s, and moved `replicas` inline.

~~Note: Matt and I are still playing around with whether we want to keep the `tsconfig.json` in the templates or not. If we decide we don't want it, I'll remove it in a follow-up PR.~~ We're going to keep the `tsconfig.json` for now.

---

If you want to try it out before this PR is merged:

```bash
pulumi new https://github.com/pulumi/templates/tree/justin/k8s/kubernetes-typescript
```

After it's merged it will be:

```bash
pulumi new kubernetes-typescript
```

### DIFF
--- a/kubernetes-typescript/index.ts
+++ b/kubernetes-typescript/index.ts
@@ -1,0 +1,20 @@
+import * as pulumi from "@pulumi/pulumi";
+import * as k8s from "@pulumi/kubernetes";
+
+const name = pulumi.getProject();
+
+const config = new pulumi.Config("kubeapp");
+const image = config.require("image");
+const replicas = config.getNumber("replicas") || 1;
+
+const appLabels = { app: name };
+const deployment = new k8s.apps.v1beta1.Deployment(name, {
+    spec: {
+        selector: { matchLabels: appLabels },
+        replicas: replicas,
+        template: {
+            metadata: { labels: appLabels },
+            spec: { containers: [{ name: name, image: image }] }
+        }
+    }
+});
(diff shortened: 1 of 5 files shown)


kubernetes-typescript/index.ts
```diff
@@ -0,0 +1,20 @@
+import * as pulumi from "@pulumi/pulumi";
+import * as k8s from "@pulumi/kubernetes";
+
+const name = pulumi.getProject();
+
+const config = new pulumi.Config("kubeapp");
```
justinvp: This will be changed to `const config = new pulumi.Config();` and `kubeapp:image` in `Pulumi.yaml` will be changed to just `image` after https://github.com/pulumi/pulumi/issues/1768 lands and is released.

joeduffy: I'm a little confused by this; shouldn't it be `const config = new pulumi.Config()`, so that it simply inherits the project name?

joeduffy:
Do you also plan to add a `kubernetes-javascript` template?

kubernetes-typescript/index.ts
justinvp: Yes, it *should* be `const config = new pulumi.Config()`, but unfortunately it can't until https://github.com/pulumi/pulumi/issues/1768 is fixed.

justinvp:
> Do you also plan to add a `kubernetes-javascript` template?

Yup. Once we're happy with this, I'll add the javascript template. It's on the list in #34.

kubernetes-typescript/index.ts
joeduffy: Oh, gotcha, so `new pulumi.Config()` effectively doesn't work as-is?

PR #37, opened 2018-09-02
Upgrade from Kubernetes v1beta1 to v1

mingfang:


joeduffy:
Looks good to me, thanks for doing this!

We should also update the JavaScript template: https://github.com/pulumi/templates/blob/master/kubernetes-javascript/index.js

@hausdorff @justinvp Am curious, any reason we had been using `v1beta1`? I've noticed this in a few examples now, including sometimes mixing `v1.Service` with `v1beta1.Deployment`. Is this intentional?

justinvp:
Wasn't intentional from me. @hausdorff?

hausdorff:
@joeduffy @justinvp The main reason is that we are running an old version of minikube in our CI. I don't think we're running these templates through a CI, so my guess is that this is a safe change to make here -- we probably do want to keep close pace with recent versions of Kubernetes though so let me file an issue to follow up changing the CI.

justinvp:
Thanks, @mingfang! Would you like to submit a PR to update the JavaScript template as well?

mingfang:
@justinvp Yes I will work on it.

PR #40, opened 2018-09-25
Stop replacing the project name in package.json

justinvp:
We've moved away from using the ${PROJECT} and ${DESCRIPTION} replacement strings in Pulumi.yaml and package.json, primarily so that `pulumi new <url>` and `pulumi up <url>` can be used with examples that don't use these replacement strings (allowing such examples to be usable/buildable on their own without any templating).

Instead of doing a search/replace of these replacement strings, v0.15.2 of the CLI now overwrites any existing `name` and `description` values in Pulumi.yaml and saves the file, and doesn't do anything to update/save the `name` specified in package.json. The replacement strings can be used, but they are no longer required with this version of the CLI.

However, v0.15.0 of the CLI only does a search/replace, so we must keep ${PROJECT} and ${DESCRIPTION} in Pulumi.yaml in the templates in pulumi/templates as long as we want v0.15.0 to continue working with our templates. (Earlier versions of the CLI before v0.15.0 download templates from S3, and therefore aren't affected by any changes we make to templates in this repo (unless we force updates to the templates in S3)).

But there's no need to keep ${PROJECT} in package.json. Since we decided we're no longer going to change the project name in package.json, we can simply remove the replacement string there.

This works around an issue where we currently allow project names that can have characters (such as `/`) that NPM doesn't allow for the `name` in package.json.

Fixes https://github.com/pulumi/pulumi/issues/1979

### DIFF
--- a/aws-javascript/package.json
+++ b/aws-javascript/package.json
@@ -1,5 +1,5 @@
 {
-    "name": "${PROJECT}",
+    "name": "aws-javascript",
     "main": "index.js",
     "dependencies": {
         "@pulumi/pulumi": "latest",
--- a/aws-typescript/package.json
+++ b/aws-typescript/package.json
@@ -1,5 +1,5 @@
 {
-    "name": "${PROJECT}",
+    "name": "aws-typescript",
     "devDependencies": {
         "@types/node": "latest"
     },
--- a/azure-javascript/package.json
+++ b/azure-javascript/package.json
@@ -1,5 +1,5 @@
 {
-    "name": "${PROJECT}",
+    "name": "azure-javascript",
     "main": "index.js",
     "dependencies": {
         "@pulumi/pulumi": "latest",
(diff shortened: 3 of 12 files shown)


aws-javascript/package.json
```diff
@@ -1,5 +1,5 @@
 {
-    "name": "${PROJECT}",
+    "name": "pulumi-project",
```
justinvp: I went back and forth whether these should use the name of the template, e.g. `aws-javascript` or just use the same generic value (e.g. `pulumi-project` or just `project`) across all templates. I don't have a strong preference either way. @lukehoban, do you have a preference?

This value doesn't really matter that much unless you're creating a library that you want to distribute on npm.

justinvp: Fixed merge conflict and decided to use the template names instead of generic `pulumi-project`. Pushed updated commit.

justinvp:
@lukehoban, mind taking a look when you have a chance?

justinvp:
I'm going to go ahead and merge as a workaround for https://github.com/pulumi/pulumi/issues/1979. We can always tweak/revert as needed.

PR #41, opened 2018-09-29
Export the Kubernetes deployment name

joeduffy:
This change exports the resulting Kubernetes deployment name
from the templates, making it easier to tool. E.g.

    $ pulumi new kubernetes-typescript
    ...
    $ kubectl get deployment $(pulumi stack output name)

### DIFF
--- a/kubernetes-javascript/index.js
+++ b/kubernetes-javascript/index.js
@@ -12,3 +12,4 @@ const deployment = new k8s.apps.v1.Deployment("nginx", {
         }
     }
 });
+exports.name = deployment.metadata.apply(m => m.name);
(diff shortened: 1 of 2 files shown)


kubernetes-javascript/index.js
```diff
@@ -12,3 +12,4 @@ const deployment = new k8s.apps.v1.Deployment("nginx", {
         }
     }
 });
+export const name = deployment.metadata.apply(m => m.name);
```
lukehoban: This won’t work in JavaScript. We use `export.name = ` instead.

joeduffy: Ah right, whoops.

joeduffy: Well, technically, it _does_ work in modern JavaScript. But to be consistent, I'll change to `exports.name`.

lukehoban: Unfortunately, Node.js, which is the host runtime for Pulumi, does not support this in `.js` files and has no plans to. 

https://nodejs.org/api/esm.html

Issue #42, opened 2018-11-13
Update Python templates as needed


justinvp:
Ensure the Python templates work with the latest bits and bump versions in requirements.txt as needed.

PR #43, opened 2018-11-14
Update Python templates

swgillespie:
Now that Python 3 support has landed, we need a few minor tweaks to the templates to get them to work:

1. `pulumi.output` was renamed to `pulumi.export` to not collide with the `Output` type and module, so we must rename the uses of `pulumi.output` in each template
2. I bumped the minimum version of each dependency specified in `requirements.txt` to be the first Python 3 SDK version.

I'm not sure how to test these examples short of manually testing them out.


swgillespie:
Fixes https://github.com/pulumi/templates/issues/42

PR #44, opened 2018-11-21
Don't use "dev" tags in templates

ellismg:
Having `pulumi new` generate a project that just installs daily builds
is likely to lead to a bad experience.

Move to "latest" like the other templates, which picks up the most
recently released version.

### DIFF
--- a/hello-aws-javascript/package.json
+++ b/hello-aws-javascript/package.json
@@ -2,7 +2,7 @@
     "name": "hello-aws-javascript",
     "main": "index.js",
     "dependencies": {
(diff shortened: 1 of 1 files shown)


ellismg:
`git grep '"dev"'` claims this was the only template with an issue.

justinvp:
I see that some of our examples in the `pulumi/examples` repo have "dev" dependencies. Should those all move to "latest"?

lukehoban:
> I see that some of our examples in the pulumi/examples repo have "dev" dependencies. Should those all move to "latest"?

Yes.

PR #48, opened 2019-02-27
Prefer dash over underscore for Python packages

justinvp:
In #47, @lblackstone suggested using `pulumi-kubernetes` over `pulumi_kubernetes`, so I assume we want to be consistent across all our Python templates.

Also, bump versions while making changes here.

As a follow-up, it'd be nice to update the README.md for each package to use the dash over the underscore, e.g. https://github.com/pulumi/pulumi-aws/blob/master/README.md#python, and any other references to `pip install` in the docs.


lukehoban:
> As a follow-up, it'd be nice to update the README.md for each package to use the dash over the underscore, e.g. https://github.com/pulumi/pulumi-aws/blob/master/README.md#python, and any other references to `pip install` in the docs.

Can we open an issue to track this? I prefer consistency over any particular style, so if we aren't committed to fixing this everywhere "now", I'd honestly prefer not to make any change here.

justinvp:
> Can we open an issue to track this? I prefer consistency over any particular style, so if we aren't committed to fixing this everywhere "now", I'd honestly prefer not to make any change here.

My plan was to open PRs everywhere else "now" after this was reviewed/merged.

justinvp:
Closing stale PR

PR #53, opened 2019-05-09
add linode to templates

displague:
Add Linode to the templates.

These Linode templates operate on Linode Instances and offer the minimal required arguments.

In practice, some known rootPass, authorizedKeys, or authorizedUsers values should be provided.


linode-go/Gopkg.toml
```diff
@@ -0,0 +1,11 @@
+[[constraint]]
+  version = "v0.18.0"
```
justinvp:
```suggestion
  version = "v0.17.10"
```

linode-typescript/Pulumi.yaml
```diff
@@ -0,0 +1,8 @@
+name: ${PROJECT}
+description: ${DESCRIPTION}
+runtime: nodejs
+template:
+  description: A minimal Linode TypeScript Pulumi program
+  config:
+    linode:project:
```
justinvp: `linode:project` doesn't look like a config value that Linode projects would need to set.

Looking at the config values for the Linode provider, I see [`linode:token`](https://github.com/pulumi/pulumi-linode/blob/5a62d364955166cc43bfecd9f5791089bd281d44/sdk/nodejs/config/vars.ts#L12) is _potentially_ the only required config value that Linode projects would need to set to be functional, though, I see it could also be specified as an environment variable.

It might be worth specifying it as a secret config value here to advertise it as a required config value:

```yaml
name: ${PROJECT}
description: ${DESCRIPTION}
runtime: nodejs
template:
  description: A minimal Linode TypeScript Pulumi program
  config:
    linode:token:
      description: The token that allows you access to your Linode account
      secret: true
```

Same applies for all the other languages

linode-python/Pulumi.yaml
```diff
@@ -0,0 +1,8 @@
+name: ${PROJECT}
+description: ${DESCRIPTION}
+runtime: python
+template:
+  description: A minimal Linode Python Pulumi program
+  config:
+    gcp:project:
```
justinvp: Replace with `linode:token`. See my comment for `linode-typescript/Pulumi.yaml`.

linode-go/Pulumi.yaml
```diff
@@ -0,0 +1,8 @@
+name: ${PROJECT}
+description: ${DESCRIPTION}
+runtime: go
+template:
+  description: A minimal Linode Go Pulumi program
+  config:
+    gcp:project:
```
justinvp: Replace with `linode:token`. See my comment for `linode-typescript/Pulumi.yaml`.

linode-javascript/Pulumi.yaml
```diff
@@ -0,0 +1,8 @@
+name: ${PROJECT}
+description: ${DESCRIPTION}
+runtime: nodejs
+template:
+  description: A minimal Linode JavaScript Pulumi program
+  config:
+    linode:project:
```
justinvp: Replace with `linode:token`. See my comment for `linode-typescript/Pulumi.yaml`.

PR #56, opened 2019-06-14
No need to pass the location in Azure templates

justinvp:
The location is flowed automatically now.


stack72:
LGTM! Nice change here - keeps us in line with the real flow

PR #57, opened 2019-06-18
Addition of templates for DigitalOcean

stack72:


jen20:
Just opened #58 since this reminded me.

stack72:
Spoke to @jen20 about this offline and we can merge these now and come back to #58 at a later time

PR #61, opened 2019-08-13
Pin NPM dependencies to `0.x.y`

lukehoban:
Until we have finalized the preparations for `1.0.0-beta.1` across providers, we need to ensure that we fix this so that users don't get a mix of Pulumi package version.


lukehoban:
Note that we plan to undo this (or update to pin to `^1.0.0`) after relaxing some checks in the CLI.

PR #65, opened 2019-09-04
Upgrade all of the core packages to 1.0.0 dependencies

stack72:


aws-typescript/package.json
```diff
@@ -4,8 +4,8 @@
         "@types/node": "latest"
     },
     "dependencies": {
-        "@pulumi/pulumi": "^0.17.28",
-        "@pulumi/aws": "^0.18.27",
+        "@pulumi/pulumi": "^1.0.0",
+        "@pulumi/aws": "^1.0.0",
         "@pulumi/awsx": "^0.18.8"
```
lukehoban: Nit - should this be ^0.18.10 like in the js version of this template?

stack72:
@lukehoban I will follow up for the other providers once we have discussed those providers today

aws-typescript/package.json
stack72: Fixed - nice catch!

PR #77, opened 2019-12-10
Add Kubernetes templates for C# and F#

lukehoban:


kubernetes-fsharp/Program.fs
```diff
@@ -0,0 +1,47 @@
...
+                      inputList [
+                        input (
+                          ContainerArgs
+                            (Name = input "nginx",
+                             Image = input "nginx",
+                             Ports = 
+                              inputList [
+                                input (
+                                 ContainerPortArgs
+                                   (ContainerPortValue = input 80))]))]))))))))
```
lukehoban: @mikhailshilkov Would love your feedback on whether there is a better way to format this

lukehoban:
Merging so these templates are available - but happy to follow up if there is any feedback on formatting changes we should make to these.

PR #80, opened 2020-01-21
Adding ability to run tests against the templates

stack72:


justinvp:
@stack72, LMK if/when this is ready for review.

stack72:
@justinvp this is ready for review - as we start to add the credentials, we can release the strictness of the conditions. But I will add a followup issue to track the additions of the other templates

tests/template_test.go
```diff
@@ -0,0 +1,119 @@
...
+
+			stdOut, stdErr := e.RunCommand("pulumi", "new", templateName, "-f", "-s", "template-test")
+
+			path, err := workspace.DetectProjectPathFrom(e.RootPath)
+			assert.NoError(t, err)
+			assert.NotEmpty(t, path)
+
+			if path == "" {
+				panic(fmt.Sprintf("path is empty: %v: %v", stdOut, stdErr))
+			}
```
justinvp: Why the `if path == ""` check? Aren't we asserting `path` isn't empty on line 87?

stack72: great catch! this was debugging

PR #84, opened 2020-02-09
Refactor template tests to be able to pass in a whitelisted set of tests

stack72:


tests/template_test.go
```diff
@@ -113,3 +120,13 @@ func deleteIfNotFailed(e *ptesting.Environment) {
 		e.DeleteEnvironment()
 	}
 }
+
+func isWhiteListedTest(templateName string, whiteListPrefix []string) bool {
+	for _, whiteList := range whiteListPrefix {
+		if strings.Contains(templateName, whiteList) {
```
mikhailshilkov: Since you check on `Contains`, "prefix" sounds like a confusing name to me (it would be good for Starts With)

tests/template_test.go
```diff
@@ -67,11 +72,13 @@ func TestTemplates(t *testing.T) {
...
 	assert.NoError(t, err)
 
+	whiteListedTests := strings.Split(testPrefixWhiteList, ",")
+
 	for _, template := range templates {
 		templateName := template.Name
 		t.Run(templateName, func(t *testing.T) {
 			t.Parallel()
-			if !strings.Contains(templateName, "aws") || strings.Contains(templateName, "go") {
+			if !isWhiteListedTest(templateName, whiteListedTests) || strings.Contains(templateName, "go") {
```
mikhailshilkov: Your "rule engine" covers only 1 exclusion out of the current two. Can we come up with the one that would cover both needs?

tests/template_test.go
```diff
@@ -13,37 +13,42 @@ import (
 )
 
 func TestTemplates(t *testing.T) {
+	testPrefixWhiteList := os.Getenv("TEST_PREFIX_WHITE_LIST")
```
mikhailshilkov: Should we black-list instead of white-listing?

stack72:
Refactored this to be a blacklist scenario :) This is a temporary thing for now which will help us at least achieve all coverage

tests/template_test.go
```diff
@@ -13,37 +13,42 @@ import (
 )
 
 func TestTemplates(t *testing.T) {
+	blackListedTests := os.Getenv("BLACK_LISTED_TESTS")
+	if blackListedTests == "" {
+		t.Skip("Skipping tests due to missing `BLACK_LISTED_TESTS` env var")
```
mikhailshilkov: Alternatively, we could assume it to be empty

stack72: yeah i am fixing that right now :)

PR #87, opened 2020-02-16
Enable Go tests

stack72:


tests/template_test.go
```diff
@@ -71,7 +71,7 @@ func TestTemplates(t *testing.T) {
 		templateName := template.Name
 		t.Run(templateName, func(t *testing.T) {
 			t.Parallel()
-			if !strings.Contains(templateName, "aws") || strings.Contains(templateName, "go") {
+			if !strings.Contains(templateName, "aws") {
```
lukehoban: Why are we skipping AWS tests?

stack72: We are not - we are skipping everything BUT AWS Tests as we move these to run from this repo - I am immediately following up with other tests being added atm :)

PR #89, opened 2020-02-17
Adding the GCP tests to the provider runs

stack72:


gcp-csharp/${PROJECT}.csproj
```diff
@@ -7,7 +7,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Pulumi.Gcp" Version="1.9.0-preview" />
+    <PackageReference Include="Pulumi.Gcp" Version="2.6.0-preview" />
```
mikhailshilkov: Possibly unrelated, but our examples still mostly use 1.x. Should we start upgrading those?

PR #91, opened 2020-02-17
Migrate to GitHub Actions

stack72:


jen20:
A few comments on this:

- It looks great and is good to merge as-is IMO, with the single change of adding a workflow name so it doesn’t show up in actions as the file path.
- The actions.yml is _very_ complex and we should factor out as much of that as possible into separate actions which are independently testable at some point in the future. It would be nice if the GitHub Actions team would support imports here in future in the same way as Azure DevOps does.
- +1000 for getting rid of the credentials file in the repo...

stack72:
@jen20 yeah this a first step - a LOT more work will happen in this area :) This gives us the immediate step of cross platform template tests which we didn't have before

PR #92, opened 2020-02-17
Bump the versions of the dependencies + stop replacing ${PROJECT} in name

stack72:
Fixes: #76


jen20:
We should probably not have a Gopkg.toml anymore?

stack72:
@jen20 the Go templates have not been migrated away from dep yet so this is something that's necessary right now :/

PR #98, opened 2020-02-20
Fix a minor issue with the gcp-go template

praneetloke:


EvanBoyle:
@praneetloke What version of the CLI and gcp provider `$ pulumi plugin ls` are you using?

This property should be a `pulumi.StringOutput` which doesn't need a function call. 

Screen Shot 2020-02-20 at 12 07 14 PM

praneetloke:
Oh that's interesting. Here's what I see:

Screen Shot 2020-02-20 at 12 46 28

All I did was `pulumi new` and then picked the `gcp-go` template. My Pulumi CLI is 1.10.1 and the GCP plugin is 1.0.0, so that could be it then? I guess I might have to run a `dep ensure` first to get the latest?

EvanBoyle:
try `pulumi plugin install resource gcp 2.6.0 && dep ensure`

praneetloke:
> pulumi plugin install resource gcp 2.6.0 && dep ensure

OK that worked. Is there something we should do here to ensure that someone else doesn't hit this? Like running `dep ensure` as part of `pulumi new`? It's sort of synonymous to the CLI running `npm install` (or `yarn`) for the nodejs counterparts. I'll close this PR.

praneetloke:
Also, I am assuming that `dep ensure` would have installed the required plugin version. But perhaps that is not possible with `dep`?

EvanBoyle:
These steps are also listed in the getting started guides, so we're doing
the best we can for now. More or less a bandaid until we land the other
work.

praneetloke:
> These steps are also listed in the getting started guides

Got it. I didn't start from that flow. I went straight to `pulumi new` and chose the new template. It would be nice to see the same message as in our Get Started flow, in the CLI's post `new` instructions as well, if that is something that's easy/possible to do.

PR #100, opened 2020-03-10
Upgrade Azure templates to 2.0.0

lukehoban:


mikhailshilkov:
Do you mind updating F# and VB too or shall I do that?

lukehoban:
> Do you mind updating F# and VB too or shall I do that?

Ahh - sorry about missing those - I'll submit a follow-up PR with those.

PR #105, opened 2020-03-30
Update go and aws-go to use go modules

EvanBoyle:


stack72:
@EvanBoyle this LGTM but I think the docs will need updated as well right? I mean, right now we say to use dep ensure?

EvanBoyle:
@stack72 we'll need to remove this line once we update everything: https://github.com/pulumi/pulumi/blob/d3f5bbce48f8d3830ddaedffaa702ee9d8a28ebc/pkg/cmd/new.go#L681

stack72:
well, not remove it - we will need to make it `go mod download` right?

EvanBoyle:
@stack72 both `go build` and `go run` will automatically acquire dependencies. I tested it this morning. Here's a PR to remove that message: https://github.com/pulumi/pulumi/pull/4232

PR #106, opened 2020-03-31
Update GCP & Azure templates to go modules 

EvanBoyle:
Did a pulumi preview on each to validate.


EvanBoyle:
fixes https://github.com/pulumi/pulumi/issues/3817

Issue #107, opened 2020-04-13
Update all dependencies for v2.0 release


jaxxstorm:
With the 2.0 release, we'll need to update all node, python, Go and dotnet dependencies for the 2.0 release.

All packages will get a major version bump, so we probably need to automate this.

We also need to make sure all deprecated package management systems have been removed, such as Gopkg

Issue #104, opened 2020-03-25
Add Go and .NET templates to the default set


joeduffy:
We filter out all non-TS/JS/Python templates by default. After Go and .NET are GA, we should probably add those to the default set of major cloud templates we show by default.

PR #108, opened 2020-04-13
Prepare templates repository for v2.0.0 launch

stack72:
Fixes: #107
Fixes: #104 - please note that this only for CSharp and Go and excludes FSharp and VB
Fixes: https://github.com/pulumi/pulumi/issues/4216
 
This also adds a kubernetes-go template

Related: https://github.com/pulumi/home/issues/746

The reason this is in draft is that we want to have the v2.0.0 dependency in the go.mod for the testing.go file - this can't be done until the release has been cut


justinvp:
Also bump `"@types/node": "^8.0.0"` to `"@types/node": "^10.0.0"` given https://github.com/pulumi/docs/pull/2724?

stack72:
> Also bump `"@types/node": "^8.0.0"` to `"@types/node": "^10.0.0"` given [pulumi/docs#2724](https://github.com/pulumi/docs/pull/2724)?

Any reason we would choose 10.0.0 and not go further up the list to 13.x.x?

justinvp:
> Any reason we would choose 10.0.0 and not go further up the list to 13.x.x?

10 is the minimum Active release we support and is what @lukehoban chose [here](https://github.com/pulumi/docs/pull/2724/files#diff-03325a6302d74f8cc58805a36a1793ec):

```diff
-        "@types/node": "^8.0.26"
+        "@types/node": "^10.0.0"
```

I'll defer to him.

lukehoban:
Yeah - sticking with `^10.0.0` makes sense to me for now.  That is the minimum we support, and so will pick up type definitions that are guaranteed to work for any Pulumi user.  Folks who are on newer versions can manually update this to depend on more recent Node.js features (though that is fairly rare).

alicloud-csharp/${PROJECT}.csproj
```diff
@@ -7,7 +7,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Pulumi.Alicloud" Version="1.5.0-preview" />
+    <PackageReference Include="Pulumi.Alicloud" Version="2.0.0" />
```
mikhailshilkov, line 10: I think we want to use floating versions like `2.*` to avoid changing templates frequently

stack72:
@justinvp / @lukehoban / @mikhailshilkov feedback acted upon

PR #113, opened 2020-04-21
fix gcp version in go.mod

EvanBoyle:


gcp-go/go.mod
```diff
@@ -3,6 +3,6 @@ module ${PROJECT}
 go 1.13
 
 require (
-	github.com/pulumi/pulumi-gcp/sdk/v3 v3.0.0
+	github.com/pulumi/pulumi-gcp/sdk/v3 v3.0.1
```
lukehoban, line 6: Curious - what's our plan for keeping these "up to date"?  I don't think we're going to want to manually bump for every patch release in every provider.

EvanBoyle: No plan at the moment. This is a good callout, I created an issue to track: https://github.com/pulumi/templates/issues/114

PR #118, opened 2020-05-08
Bump awsx dependencies in AWS JS/TS templates

leezen:


lukehoban:
I think there is also an updated needed in `hello-aws-javascript/package.json`?

leezen:
Ah, thanks. I missed that. Just confirming via grep, there's nothing else:

```
aws-javascript/package.json:        "@pulumi/awsx": "^0.20.0"
aws-typescript/package.json:        "@pulumi/awsx": "^0.20.0"
hello-aws-javascript/package.json:        "@pulumi/awsx": "^0.18.10"
```

PR #125, opened 2020-06-24
Temporarily exclude grpcio v1.30.0, which breaks Pulumi Python programs

justinvp:
Part of https://github.com/pulumi/pulumi/issues/4881

Once we've released a new version of `pulumi` package that has this constraint, we can revert this commit.


justinvp:
You can test this out, if you'd like:

```
$ pulumi new https://github.com/pulumi/templates/tree/justin/grpcio
```

PR #159, opened 2021-02-23
Run the templates from the PR source if exists

stack72:
If we are changing the templates in a PR, then we should pass
the ref to the template checkout otherwise we are running tests
from the main line not the PR itself


mikhailshilkov (approved):
Thanks a ton!

PR #162, opened 2021-03-04
Remove deprecated 'latest' from Azure native templates

davidwrede:
After making Azure-native the default, some of the current templates still have references to the deprecated `latest` namespace/module, so this updates the templates to use the top-level namespace/module.


mikhailshilkov (approved):
🤦 how did I miss these... thank you!

PR #180, opened 2021-06-16
WIP on perf metric collector

t0yv0:


t0yv0:
Checking it in so I can start iterating on running it.

PR #183, opened 2021-06-24
Finish cli perf metric capture from template tests

t0yv0:


t0yv0:
@stack72 PTAL - it's basically the same structure as in pulumi/examples. It seems to be working great for me - I get the data I was after. I didn't modify the blacklists - kept the original blacklists. Seems okay?

The one difference affecting mainline here is I've upgraded go.mod references to pulumi/{pkg, sdk} to latest.

t0yv0:
@stack72 need a ✔️ here or pls let me know if I should check minor tweaks in w/o review.

stack72 (approved):

PR #190, opened 2021-07-20
Create civo-go template for pulumi new

dirien:

Hi good folks at Pulumi,

It would be nice to have a civo template to use in pulumi new

i start with the civo-go template.


stack72:
Thanks for this @dirien 

This LGTM! I will merge it and then fix up our CI process around this

PR #195, opened 2021-08-03
Add aiven-go template to pulumi

dirien:

Hi nice folks at Pulumi,

It would be sweet to have a Aiven template to use in pulumi cli.

So I created this template.


stack72:
thanks so much for this @dirien 

I merged this locally so I could exclude from our CI runs for now

PR #213, opened 2021-09-21
chore: bump from node v10 to v14

cobraz:
We probably want to use Node 14, as it's recommended for most applications. Node v10 is EOL.


stack72:

[1]: https://github.com/pulumi/templates/actions/runs/1339859652

PR #215, opened 2021-09-22
Upgrade to Go 1.16.x

stack72:


emiliza:
As part of upgrading, should we update all the go.mod files in the subdirectories?

stack72:
> As part of upgrading, should we update all the go.mod files in the subdirectories?

I don't think that's required as we actually have the Go versions in the corresponding SDKs as the lower variants... but I may be wrong here

PR #221, opened 2021-10-10
feat: add civo-javascript, civo-typescript and civo-python

dirien:

This PR adds the missing civo templates for javascript, typescript and python.


stack72:
Thanks so much for these @dirien!

PR #224, opened 2021-10-29
chore: bump from node v14 to v16

simenandre:


stack72:

[1]: https://github.com/pulumi/templates/actions/runs/1413747417

cobraz:
Should we get this landed soon? Node v14 is now in maintainance, and node v18 will be LTS in late august.

I think we should try to update so templates follow the latest LTS, this pull request will do that.

cnunciato:
@cobraz Sorry we didn't get to this sooner! We'll be bumping all Node templates past 16 shortly, so we should probably close this.

PR #238, opened 2021-12-27
feat: add auth0 templates (golang, type- and javascript)

dirien:
feat: add auth0 templates (golang, type- and javascript)


stack72:
Superceded via #240

PR #247, opened 2022-01-31
Upate @pulumi/gcp version

cstanger:
in oder to use the lates version of the GCP Pulumi provider the package.json is updated to use the version ^6.0.0


stack72 (approved):
Thank you so much! Apologies this wasn't updated

PR #260, opened 2022-04-04
chore: update civo version to 2.2.0 and fix breaking changes

dirien:
chore: update civo version to 2.2.0 and fix breaking changes


dirien:
Thanks @stack72 for the fix with the pulumi-civo go provider! That helped to add the go part to this PR too! ❤️

kunal-kushwaha:
Thank You!

PR #267, opened 2022-04-21
feat: digitalocean:token as config in Pulumi.yaml

dirien:
Hi,

This PR adds the digitalocean:token as config into the Pulumi.yaml.

I used the template the last times quite often und had to set the token via `pulumi config...`


dirien:
You are right @stack72, users should choose what fits them the best. Going to close the PR.

PR #273, opened 2022-05-04
Add new languages

mikhailshilkov:


alicloud-yaml/Pulumi.yaml
```diff
@@ -0,0 +1,18 @@
...
+description: ${DESCRIPTION}
+runtime: yaml
+template:
+  description: A minimal AliCloud Pulumi YAML program
+  config:
+    alicloud:region:
+      description: The AliCloud region to deploy into
+      default: cn-beijing
+
+resources:
```
mikhailshilkov, line 11: Is it intended that we have `resources` in both files? Will it break folks if we ever fix templating in a future CLI version? Why have it here?

Frassle: I'd guess the intention is if we fix templating we'll just start using this file and ignore the append one

mikhailshilkov: hmm... kk, I guess it works

PR #280, opened 2022-05-06
Add `location` to `azure-java` template config

cnunciato:
Without this, we get the following first-run experience:

```
Diagnostics:
  azure-native:resources:ResourceGroup (resourceGroup):
    error: azure-native:resources:ResourceGroup resource 'resourceGroup' has a problem: missing required property 'location'. Either set it explicitly or configure it with 'pulumi config set azure-native:location <value>'.
```

### DIFF
--- a/azure-java/Pulumi.yaml
+++ b/azure-java/Pulumi.yaml
@@ -2,7 +2,7 @@ name: ${PROJECT}
 description: ${DESCRIPTION}
 runtime: java
 template:
-  description: "A minimal Azure Native Java Pulumi program"
+  description: A minimal Azure Native Java Pulumi program
   important: true
   config:
     azure-native:location:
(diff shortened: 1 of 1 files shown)


azure-java/Pulumi.yaml
```diff
@@ -2,4 +2,9 @@ name: ${PROJECT}
 description: ${DESCRIPTION}
 runtime: java
 template:
-  description: A minimal Azure Native Java Pulumi program
+  description: "A minimal Azure Native Java Pulumi program"
```
iwahbe, line 5: Nit: I think the quotes are not idiomatic, especially if we have `config.azure-native:location.description` unquoted.
```suggestion
  description: A minimal Azure Native Java Pulumi program
```

iwahbe (approved):

dixler:
Do the other azure templates have this issue? I'd imagine it's probably an azure specific thing.

cnunciato:
> Do the other azure templates have this issue? I'd imagine it's probably an azure specific thing.

It is Azure-specific, yeah, and all of the other `azure-*` templates have it. Only this one's missing it.

PR #281, opened 2022-05-06
Change the name of the bucket resource to my-bucket

cnunciato:
Going through the recent additions to the getting-started guides, I noticed we use the resource name `bucket` in these two templates, but we use `my-bucket` in all of the others. Since we have shared code in our guides that refers to this bucket as `my-bucket`, it'd be nice to update it here, both for consistency and to keep from having to special-case a bunch of the prose in the guides. Thank you!

### DIFF
--- a/aws-yaml/Pulumi.yaml
+++ b/aws-yaml/Pulumi.yaml
@@ -11,9 +11,9 @@ template:
 
 resources:
   # Create an AWS resource (S3 Bucket)
-  bucket:
+  my-bucket:
     type: aws:s3:Bucket
 
 outputs:
   # Export the name of the bucket
-  bucketName: ${bucket.id}
+  bucketName: ${my-bucket.id}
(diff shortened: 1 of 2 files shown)


mikhailshilkov (approved):

iwahbe:
It doesn't really matter, but the reason I called it `my-bucket` instead of bucket is that other example have the variable called `bucket` but name the resource `my-bucket`. YAML combines resource name and variable name. This change makes the YAML examples look more like the others but diverges their actual behavior to be less like the other examples.

cnunciato:
Thanks — this is useful info that should make it into the docs. I wondered myself whether there was way to name something differently from its resource name. Is this documented yet? I haven’t seen it yet.

iwahbe:
Its in [the docs](https://github.com/pulumi/pulumi-yaml#resources). It's definitely a sentence that was written as part of an implementation guide, not for users. You need to know the difference between a [logical name](https://www.pulumi.com/docs/intro/concepts/resources/names/#logicalname) and a lexical name.

> The value of resources is an object whose keys are logical resource names by which the resource will be referenced in expressions within the program, and whose values which are elements of the schema below.

PR #282, opened 2022-05-07
Add `project` to `gcp-java` template config

cnunciato:
Without this, we get the following first-run experience:

```
Diagnostics:
  pulumi:pulumi:Stack (quickstart-dev):
    error: update failed
 
  gcp:storage:Bucket (my-bucket):
    error: 1 error occurred:
        * project: required field is not set
```

Looks like all other `gcp-*` templates have this as well.


dixler (approved):
LGTM

PR #283, opened 2022-05-07
Add a template for java-jbang

maxandersen:
matching with adding jbang to pulumi-language-java.

See https://github.com/pulumi/pulumi-java/pull/575

### DIFF
--- a/java-jbang/src/main.java
+++ b/java-jbang/src/main.java
@@ -1,0 +1,12 @@
+//DEPS com.pulumi:pulumi:0.1.0
+
+import com.pulumi.Pulumi;
+import com.pulumi.core.Output;
+
+public class main {
+    public static void main(String[] args) {
+        Pulumi.run(ctx -> {
+            ctx.export("exampleOutput", Output.of("example"));
+        });
+    }
+}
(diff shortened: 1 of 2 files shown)


maxandersen:
Can someone /run-templates ?

t0yv0:
Waiting on https://github.com/pulumi/pulumi-java/pull/575 to be released in the next version of pulumi.

maxandersen:
Sounds good. Just heads up I'm on PTO next week so just be aware I might not react before week after :)

maxandersen:
Now jbang support is Released anything blocking this from being merged?

t0yv0 (approved):
Tested and it works with pulumi v3.34.0

java-jbang/src/main.java
```diff
@@ -0,0 +1,12 @@
+//DEPS com.pulumi:pulumi:0.1.0
```
t0yv0, line 1: Could we make it a floating ref?

t0yv0: Perhaps `//DEPS com.pulumi:pulumi:0.+` - this seems to work for me.

stack72:

[1]: https://github.com/pulumi/templates/actions/runs/2471226344

java-jbang/src/main.java
maxandersen: yes, that syntax kinda works. probably will add "(,1.0]" syntax in future.

t0yv0:
Merged via https://github.com/pulumi/templates/pull/297 thank you!

PR #284, opened 2022-05-15
Fix bucket resource name

cnunciato:
Fixes https://github.com/pulumi/docs/issues/7508.

### DIFF
--- a/gcp-yaml/Pulumi.yaml
+++ b/gcp-yaml/Pulumi.yaml
@@ -10,11 +10,11 @@ template:
 
 resources:
   # Create a GCP resource (Storage Bucket)
-  bucket:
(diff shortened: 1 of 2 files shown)


susanev (approved):

gcp-yaml/Pulumi.yaml
```diff
@@ -10,11 +10,11 @@ template:
 
 resources:
   # Create a GCP resource (Storage Bucket)
-  bucket:
+  my-bucket:
```
cnunciato, line 13: Happened to notice these as well, so fixing them up, too. They aren't causing a problem, as the .append file renames them as `my-bucket`, but I can't think of a reason not to keep them in sync.

PR #287, opened 2022-05-27
Enable java template tests

t0yv0:
Tests Java templates against the latest CLI/Java provider. This would be useful.

Surprisingly no Java installation actions are needed.

Looks like https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-Readme.md includes Java, Maven and Gradle.


tests/template_test.go
```diff
@@ -84,7 +84,7 @@ func TestTemplates(t *testing.T) {
 		template := template
 		templateName := template.Name
 
-		if isBlackListedTest(templateName, blackListed) {
+		if isBlackListedTest(templateName, blackListed) || templateName != "java" {
```
stack72, line 87: you shouldn't need the `||`, we don't need it for go

t0yv0: Not ready for review yet - I"m messing with it to only run java tests to speed up iterating on CI. I'll remove when it passes.

mikhailshilkov (approved):
LGTM. @AaronFriel are we tracking this for YAML?

t0yv0:
Running Windows Tests: 1h 38min something is off. I'm wondering if the specific changes here are to blame like the templates are hanging on Windows.

t0yv0:
It totally does die. I'll spin up a Windows VM to investigate.

AaronFriel:
@mikhailshilkov added #289 to track

t0yv0:
Running this on a Windows VM did not find any issues, seem to work, cannot repro the hang. I'm trying to add some per-test timeouts to see if this can narrow it down in the output to see which test is "bad".

tests/template_test.go
```diff
@@ -6,13 +6,16 @@ import (
...
 	"testing"
+	"time"
 
 	"github.com/pulumi/pulumi/pkg/v3/testing/integration"
 	ptesting "github.com/pulumi/pulumi/sdk/v3/go/common/testing"
 	"github.com/pulumi/pulumi/sdk/v3/go/common/workspace"
 	"github.com/stretchr/testify/assert"
 )
 
+const testTimeout = 30 * time.Minute
```
t0yv0, line 17: Added this to help diagnosing hanging tests. Feel free to increase if needed.

PR #294, opened 2022-06-07
Update dotnet templates to net6.0

### LINKED ISSUE #292: Switch templates to .NET 6
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details

.NET 6 is the current LTS version. We should switch our templates to .NET 6 and use the latest C# features like top-level statements to make the program as nice and as simple as possible.

Zaid-Ajaj:
Fixes #292 

This PR updates all dotnet templates (csharp, fsharp, vb.net) to target net6.0 and use latest features both of the runtime as well as update old template code

### C# templates
 - Now use top-level statements to run `await Deployment.RunAsync(...lambda...)` without needing an explicit `Main` entry point nor a `Program` class. 
 - Got rid of `MyStack.cs` and now using a simple lambda provided for `Deployment.RunAsync` of the type `Func<IDictionary<string, object>>` to define output exports
 - Now use output-versioned invokes where possible: calling `.Invoke(...)` instead of calling `InvokeAsync(...)` and having to juggle `Task<T>` into `Output<T>`

### F# templates
 - Whenever possible, now make use of implicit conversion without having to explicitly call `input : 'a -> Input<'a>`
 - Same as with C#, now use output-versioned invokes where possible: calling `.Invoke(...)` instead of calling `InvokeAsync(...)` and having to juggle `Async<T>` into `Output<T>`

### VB.NET templates
 - Got rid of `MyStack.vb` and now using a simple function provided for `Deployment.RunAsync` of the type `Func(Of Dictionary(Of String, Object))` to define output exports


Zaid-Ajaj:
Testing on ubuntu-latest failed with the following: the test
```
TestTemplates/azure-fsharp
```
Failed with
```
System.ArgumentNullException: [Input] Pulumi.AzureNative.Storage.Inputs.SkuArgs.Name is required but 
was not given a value (Parameter 'Name')
```
Which is weird becaue `Name` is provided in azure-fsharp like this:
```fs
// Create an Azure Storage Account
let storageAccount =
    StorageAccount("sa",
        StorageAccountArgs
            (ResourceGroupName = resourceGroup.Name,
             Sku = SkuArgs(Name = SkuName.Standard_LRS),
             Kind = Kind.StorageV2))
```
I will investigate

mikhailshilkov:
Are you still working on Windows tests? Do you know why they are still trying to load .NET Core 3.1?

Zaid-Ajaj:
@mikhailshilkov I looked into it and couldn't find where 3.1 is coming from. This repo references only dotnet 6.0.x and the `integration.ProgramTest` from pulumi/pulumi doesn't reference it either 🤔

alicloud-csharp/Program.cs
```diff
@@ -1,7 +1,15 @@
...
 {
-    static Task<int> Main() => Deployment.RunAsync<MyStack>();
-}
+    // Create an AliCloud resource (OSS Bucket)
+    var bucket = new Bucket("my-bucket");
+    
+    // Export the name of the bucket
+    return new Dictionary<string, object?>
+    {
+        ["BucketName"] = bucket.Id
```
mikhailshilkov, line 13: Since we are switching to named outputs, I suggest we format then in loweCamelCase similar to all other languages.
```suggestion
        ["bucketName"] = bucket.Id
```

If you agree, we should change that in all templates.

alicloud-csharp/Program.cs
```diff
@@ -1,7 +1,15 @@
...
+    };
+});
```
mikhailshilkov, line 15: Nit: missing new line (check all files - there's a few of those)

alicloud-visualbasic/${PROJECT}.vbproj
```diff
@@ -2,9 +2,9 @@
 
   <PropertyGroup>
     <OutputType>Exe</OutputType>
-    <TargetFramework>netcoreapp3.1</TargetFramework>
+    <TargetFramework>net6.0</TargetFramework>
     <Nullable>enable</Nullable>
-    <RootNamespace></RootNamespace>
+    <RootNamespace>PulumiVbTemplate</RootNamespace>
```
mikhailshilkov, line 7: What's the benefit of changing this? Is that specific to VB? An empty namespace feels more natural to me than `PulumiVbTemplate` that users will need to change 100% of the time.

azure-csharp/Program.cs
```diff
@@ -1,7 +1,40 @@
...
+    {
+        ResourceGroupName = resourceGroup.Name,
+        Sku = new SkuArgs
+        {
+            Name = SkuName.Standard_LRS
+        },
+        Kind = Kind.StorageV2
+    });
+
+    var storageAccountKeys = ListStorageAccountKeys.Invoke(new ListStorageAccountKeysInvokeArgs
```
mikhailshilkov, line 23: Nice change!

gcp-fsharp/Program.fs
```diff
@@ -5,9 +5,7 @@ open Pulumi.Gcp.Storage
 
 let infra () =
   // Create a GCP resource (Storage Bucket)
-  let bucket = Bucket("my-bucket",
-                    BucketArgs
-                        (Location = input  "US"))
+  let bucket = Bucket("my-bucket", BucketArgs(Location = "US"))
```
mikhailshilkov, line 8: I'm pretty sure `input` used to be required here... Has F# changed to allow implicit conversions? Or how does that work now?

mikhailshilkov (commented):
Mostly LGTM with a few tactical questions

gcp-fsharp/Program.fs
Zaid-Ajaj: F# 6 has implicit conversions https://docs.microsoft.com/en-us/dotnet/fsharp/whats-new/fsharp-6#additional-implicit-upcast-conversions 

Though see comment above where an implicit conversion was used and compiled successfully but caused a runtime error. I want to look into that soon

Zaid-Ajaj:
ToDo:
 - [ ] Export keys should be lower-case
 - [ ] Missing new line

Working on those

stack72 (approved):
We are facing windows test anomolies - we should block this PR on those test issues - we can see the correct tests are running on linux. We will manually test on windows when we have merged this PR

PR #309, opened 2022-07-11
Update kubernetes-java template

pawelprazak:
Adjusts the `kuberentes-java` template to work after breaking changes in https://github.com/pulumi/pulumi-kubernetes/pull/2055

PR #311, opened 2022-07-14
Update pulumi-java/main from master

pawelprazak:
Merge master to pulumi-java/main


pawelprazak:
@t0yv0 dependency for https://github.com/pulumi/templates/pull/309

PR #312, opened 2022-07-19
Opt out of TypeScript in JavaScript templates

t0yv0:
Projects that do not use TypeScript still pay some TypeScript-related overheads inside Pulumi. One overhead is `import "typescript"` in nodejs costing 300ms. `pulumi-language-nodejs` currently thinks that every project is a TypeScript project, it does not autodetect; unless it is told otherwise using this option. While we could consider auto-detection, I am afraid it might break some projects out there; so the safer path in this PR is to opt out of TypeScript explicitly for all the JavaScript templates so the new projects take advantage of the related optimizations.

### DIFF
--- a/alicloud-javascript/Pulumi.yaml
+++ b/alicloud-javascript/Pulumi.yaml
@@ -1,6 +1,9 @@
 name: ${PROJECT}
 description: ${DESCRIPTION}
-runtime: nodejs
+runtime:
+  name: nodejs
+  options:
+    typescript: false
 template:
   description: A minimal AliCloud JavaScript Pulumi program
   config:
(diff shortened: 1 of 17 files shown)


t0yv0:
Related: https://github.com/pulumi/pulumi/pull/10185

t0yv0:
What do folks think, should we take this approach of editing templates or instead think about autodetecting if a project is a typescript project inside `pulumi-language-nodejs`. The key disadvantage of this PR is that any speed ups we make will not materialize for projects that simply upgrade to newer version of Pulumi, since those will still think they're typescript projects unless the user edits Pulumi.yaml.

AaronFriel (approved):
lgtm

PR #313, opened 2022-07-21
Fix csharp templates

Frassle:
Tactical fix for https://github.com/pulumi/pulumi/issues/10202


mikhailshilkov (approved):
Thank you! We need to adjust Get Started guides accordingly.

Zaid-Ajaj (approved):

PR #318, opened 2022-08-08
Add initial set of architecture templates

lukehoban:
This introduces the first two architecture templates - a pattern we expect to expand to offer more "batteries included" started projects for a set of common use cases.  This PR currently includes:
- `container-aws-{typescript,python,go,csharp,yaml}`
- `serverless-aws-{typescript,python,go,csharp,yaml}`

We expect to add additional architectures, support these existing architectures on additional clouds, and support the architecture+cloud's in this PR in all Pulumi languages.

Theses are included in the templates repo so that they can be installed directly with `pulumi new container-aws-python`.  They are not right now marked as "important", so will not show up in the default list.  But as part of delivering https://github.com/pulumi/pulumi/issues/10320, we expect to change the `pulumi new` interactive prompts to include these new architectures as part of the prompts - likely after a larger set of the templates are available.

Known TODO:
- [x] Add C# for these two architectures
- [x] Add a Kubernetes Cluster architecture template
- [x] Ensure robust testing of these templates is run in CI (including validation of runtime correctness)

For follow-up PRs:

- [ ] Add Java for these two architectures

[Design doc](https://docs.google.com/document/d/17lCbBQ49qJ6Y2sJr68lij6HvFRM6BXg_paskzraUK8M/edit#)

Part of https://github.com/pulumi/marketing/issues/456.


serverless-aws-typescript/package.json
```diff
@@ -0,0 +1,13 @@
...
+		"devDependencies": {
+				"@types/node": "^14"
+		},
+		"dependencies": {
+				"@pulumi/aws": "5.10.0",
+				"@pulumi/aws-apigateway": "^0.0.11",
+				"@pulumi/awsx": "^1.0.0-beta",
+				"@pulumi/pulumi": "^3.0.0",
+				"typescript": "^4.0.0"
+		}
```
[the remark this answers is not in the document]
cnunciato: When I run this with `pulumi up`, I get:

```
Diagnostics:
  pulumi:providers:aws (default_4_30_0):
    error: no resource plugin 'pulumi-resource-aws' found in the workspace at version v4.30.0 or on your $PATH, install the plugin using `pulumi plugin install resource aws v4.30.0`
```

`serverless-aws-python` and `-go` show the same, though -- so I'm guessing that means this is expected for now?

cnunciato (commented):
Ran through these, and the container examples all look good, but the serverless ones aren't working for me. I'm assuming that's probably known, though, given the consistent errors across them?

container-aws-go/Pulumi.yaml
```diff
@@ -0,0 +1,18 @@
+name: ${PROJECT}
+description: ${DESCRIPTION}
+runtime: go
+template:
+  description: A containerized service on AWS Fargate
+  config:
```
joeduffy, line 6: I had expected (at least) two other config variables:
1) A `imageName`, in case you already have an image to deploy. That would let you have a fully functional deployment without needing code edits - e.g., just configure it to have `nginx`.

2) A `replicas` count, so you can configure and scale up/down the service.

Maybe others as I look through but those leapt out to me.

serverless-aws-typescript/index.ts
```diff
@@ -0,0 +1,24 @@
+import * as pulumi from "@pulumi/pulumi";
+import * as aws from "@pulumi/aws";
+import * as apigateway from "@pulumi/aws-apigateway";
+
+// A Lambda function to invoke
+const fn = new aws.lambda.CallbackFunction("fn", {
+    callback: async (ev, ctx) => {
```
joeduffy, line 7: Given how often folks hork themselves using callback functions -- and this is painful to suggest because I love them -- should we have code that uses archives, like with the other languages? In addition to less horkability, it's also symmetric across examples. Perhaps you could have one of each so they can pick the style that works best?

container-aws-go/Pulumi.yaml
lukehoban: > A imageName, in case you already have an image to deploy. 

The template currently does a build and push from a local folder, so that users can configure this by changing what's there instead of hard coding an image name.  My sense is this is much more "real", and makes it much easier to say "just put your code in ./app and re-run pulumi up" - which we can't say if we have this set via a fixed image name.

> That would let you have a fully functional deployment without needing code edits - e.g., just configure it to have nginx.

Note that you do have a fully functional `nginx` deployment out of the box, because the default contents of `./app` is a `Dockerfile` with `FROM nginx`.  So to a meaningful degree - you get the best of both worlds - something that works out of the box, and is easy to "make yours" by just dropping your app code in a folder.

We could in theory support "both", but it would make the template a lot more confusing.  On balance, I prefer the approach here currently.

Though I do think it makes the stronger case for the need for READMEs.

> A replicas count, so you can configure and scale up/down the service.

A sort of meta-question is what value exactly you perceive there being in having lots of config for these templates.  I started off with more config, but realized there is fairly limited value, because the only way to edit config after the initial deployment is to make an edit to the repo, and it's just as easy to edit the code as to edit the config.  The real role of config is to make it easier to have things that will differ across different stacks deployed from the same project.  

The other factor is that it's easy to go overboard with config and just make things very complicated for little value.  My sense after trying a few versions was that it's better to be simple out of the box.

serverless-aws-typescript/index.ts
lukehoban: Maybe - though I believe the "magic" of this is one of the most exciting things we can highlight for the serverless use case in TypeScript.  (Lots of feedback supporting this was the reason for getting interested in Pulumi).  

Perhaps the README for this could be the place to note the alternative approach?

container-aws-go/go.mod
```diff
@@ -0,0 +1,59 @@
+module ${PROJECT}
+
+go 1.17
```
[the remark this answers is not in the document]
lukehoban: It's 1.17 for all other templates.  I'll leave to a separate PR to update across all templates consistently.

serverless-aws-yaml/Pulumi.yaml
```diff
@@ -0,0 +1,50 @@
+name: ${PROJECT}
```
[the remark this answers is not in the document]
lukehoban: > it looks like GetSchema is not implemented

I think you must somehow have a pretty old version of this package installed?  Is it on your path somehow?

We may need to force the version - but that would be a little unfortunate.

serverless-aws-yaml/Pulumi.yaml.append
```diff
@@ -0,0 +1,40 @@
+resources:
```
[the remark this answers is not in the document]
lukehoban: This is a sort of weird hack we need currently (it's in all other YAML-based templates).  I'd love to get rid of it.

Opened https://github.com/pulumi/pulumi-yaml/issues/303 on that.

serverless-aws-typescript/package.json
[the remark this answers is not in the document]
lukehoban: Hmm - I have no idea why it would be creating a 4.30 AWS provider for you.  The packages here all use 5.x.

Let me reverify from a clean environment.

serverless-aws-yaml/Pulumi.yaml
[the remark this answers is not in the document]
cnunciato: Ah, let me verify that — may very well be.

gthuang:
when i run container-aws-python, i run into a 503 error when i open the outputted URL

container-aws-go/main.go
```diff
@@ -0,0 +1,82 @@
...
+			Path:          pulumi.String("./app"),
+		})
+		if err != nil {
+			return err
+		}
+
+		// Deploy an ECS Service on Fargate to host the application container
+		_, err = ecsx.NewFargateService(ctx, "service", &ecsx.FargateServiceArgs{
+			Cluster: cluster.Arn,
+			TaskDefinitionArgs: &ecsx.FargateServiceTaskDefinitionArgs{
```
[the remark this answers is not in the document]
lukehoban: I've updated the go.mod version, and don't see this issue after that.

container-aws-go/go.mod
```diff
@@ -0,0 +1,59 @@
...
+
+require (
+	github.com/pulumi/pulumi-aws/sdk/v5 v5.10.0
+	github.com/pulumi/pulumi-awsx/sdk v1.0.0-testplsignore.0
```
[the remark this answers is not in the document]
lukehoban: Fixed.

serverless-aws-yaml/Pulumi.yaml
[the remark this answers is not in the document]
lukehoban: I cannot repro this, and I assume it must be due to an old `aws-apigateway` on PATH.  Resolving the comment, but if you can still reproduce, let me know.

serverless-aws-typescript/package.json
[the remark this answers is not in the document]
lukehoban: I can't repro this one either, and can't figure out what would cause you to see this unless you had a binary on path that is overriding the declared dependencies here. Going to resolve, but if you still see this - let me know.

serverless-aws-typescript/index.ts
lukehoban: I'm going to keep this as is for now, but it's something we can revisit in the future and/or add to Next Steps in the marketing pages.

lukehoban:
> when i run container-aws-python, i run into a 503 error when i open the outputted URL

This may happen for a few seconds after the stack is deployed, until the container is fully running in Fargate.  I believe we *should* be waiting for steady state, so that this would not happen at all after the update has deployed, which I think might potentially be an issue in the `awsx.FargateService` library that I will raise.

lukehoban:
> > when i run container-aws-python, i run into a 503 error when i open the outputted URL
> This may happen for a few seconds after the stack is deployed, until the container is fully running in Fargate. I believe we should be waiting for steady state, so that this would not happen at all after the update has deployed, which I think might potentially be an issue in the awsx.FargateService library that I will raise.

This was apparently due to a  bug in AWSX tracked in https://github.com/pulumi/pulumi-awsx/issues/902, and is not possible to workaround in the template.  A PR is open now to fix that, so that the template will wait for the service to be ready in the future once this new version is released.

container-aws-csharp/Pulumi.yaml
```diff
@@ -0,0 +1,18 @@
+name: ${PROJECT}
+description: ${DESCRIPTION}
+runtime: yaml
```
cnunciato, line 3:
```suggestion
runtime: dotnet
```

container-aws-yaml/Pulumi.yaml
```diff
@@ -0,0 +1,72 @@
...
+runtime: yaml
+template:
+  description: A containerized service on AWS Fargate
+  config:
+    aws:region:
+      description: The AWS region to deploy into
+      default: us-west-2
+    containerPort:
+      description: The port to expose on the container
+      default: 80
```
cnunciato, line 12: When I run `pulumi new` and then `pulumi up` with this, I get:

```
➜  container-aws-yaml pulumi up
Previewing update (dev)

View Live: https://app.pulumi.com/cnunciato/container-aws-yaml/dev/previews/9f7ba2ab-1acc-4c84-bcf8-88f59cb085d7

error: failed to discover plugin requirements: Pulumi.yaml:7,18-20: default must be a string; default must be a string, and 2 other diagnostic(s)
```

Changing the values to `"80"`, `"512"`, and `"128"` get me past it, though. Is this expected? If not, I can file for it.

container-aws-yaml/Pulumi.yaml
```diff
@@ -0,0 +1,72 @@
...
+          essential: true
+          portMappings:
+          - containerPort: ${containerPort}
+            targetGroup: ${loadbalancer.defaultTargetGroup}
+    options:
+      version: 1.0.0-beta.10
+
+outputs:
+  # The URL at which the container's HTTP endpoint will be available
+  url: http://${loadbalancer.loadBalancer.dnsName}
```
cnunciato, line 72: This line seems to be causing issues for me:

image

It _looks_ right, though, compared to others that do work, so I'm not sure why this one doesn't.

cnunciato: Actually something else must be going on here. Removing that line seems to fail in the same way. 🤔

cnunciato (commented):
Thanks for the updates! `serverless-*` looks good, and `container-*` looks good as well with the exception of YAML, which I'm still seeing issues with.

container-aws-python/__main__.py
```diff
@@ -0,0 +1,41 @@
+from pulumi import Config, Output, export
+import pulumi_aws as aws
+import pulumi_awsx as awsx
+
+config = Config()
+container_port = config.get_int("containerPort") or 80
+cpu = config.get_int("cpu") or 512
+memory = config.get_int("memory") or 128
```
nimbinatus, line 8: May I suggest using the built-in default value option, instead? It's a slight bit more idiomatic to match the built-in dictionary call:

```suggestion
container_port = config.get_int("containerPort", 80)
cpu = config.get_int("cpu", 512)
memory = config.get_int("memory", 128)
```

container-aws-python/__main__.py
```diff
@@ -0,0 +1,41 @@
...
+cluster = aws.ecs.Cluster("cluster")
+
+# An ALB to serve the container endpoint to the internet
+loadbalancer = awsx.lb.ApplicationLoadBalancer("loadbalancer")
+
+# An ECR repository to store our application's container image
+repo = awsx.ecr.Repository("repo")
+
+# Build and publish our application's container image from ./app to the ECR repository
+image = awsx.ecr.Image("image",
```
nimbinatus, line 20: General PEP8 nit: There will be warnings on opening this in an editor with checks turned on as indentation for lines 21-22 and 26-38 are not visually aligned to the first argument. Two options for fixing this one:
* Option 1: Move the first argument to a new line:
```python
image = awsx.ecr.Image(
    "image",
    repository_url=repo.url,
    path="./app"
)
```
* Option 2: Move everything over to visually indent:
```python
image = awsx.ecr.Image("image",
                       repository_url=repo.url,
                       path="./app")
```
Both options are equally valid and idiomatic. I prefer the first as I'm less likely to hit line length linters and I find it more readable as things get more complicated, but that's a personal preference.

container-aws-csharp/Pulumi.yaml
```diff
@@ -0,0 +1,18 @@
...
+template:
+  description: A containerized service on AWS Fargate
+  config:
+    aws:region:
+      description: The AWS region to deploy into
+      default: us-west-2
+    containerPort:
+      description: The port to expose on the container
+      default: 80
```
nimbinatus, line 12: Out of curiosity, is there a reason these values aren't explicitly in quotes like for the YAML in the other languages?

PR #323, opened 2022-08-14
kubernetes-kotlin template

luistrigueiros:
Created kubernetes kotlin template using gradle with kotlin dsl


kubernetes-kotlin/src/main/kotlin/myproject/Support.kt
```diff
@@ -0,0 +1,55 @@
+package myproject
+
+import com.pulumi.kubernetes.apps_v1.Deployment
+import com.pulumi.kubernetes.apps_v1.DeploymentArgs
+import com.pulumi.kubernetes.apps_v1.inputs.DeploymentSpecArgs
```
t0yv0, line 5: I believe more recent versions of com.pulumi:kubernetes upgraded to the following style of imports:

```
import com.pulumi.kubernetes.apps.v1.Deployment
```

kubernetes-kotlin/gradlew
```diff
@@ -0,0 +1,234 @@
+#!/bin/sh
```
t0yv0, line 1: Pulumi should support running preinstalled `gradle` if the `./gradlew` and `gradle/wrapper` files are omitted. We currently omit them in the `gradle` template. I'm curious what is the better default for users, do you believe including gradle wrappers is the better default (in which case we should consider adding that to the plain `gradle` template also?

t0yv0:
Thank you for this contribution! I've left a few questions but we'll be happy to accept once we work through those.

It's pretty exciting to have a Kotlin template, even if the APIs we have may be skewing toward idiomatic Java and not idiomatic Kotlin. 

Note we have a general issue here that is receiving upvotes and may help guide the roadmap:

https://github.com/pulumi/pulumi-java/issues/544

Would love to hear of how you're using Kotlin with what Pulumi can do to make that experience better.

kubernetes-kotlin/gradlew
luistrigueiros: yes including the gradlew if is consider better practice in gradle because it allows not to make assumptions about what the user has or not available in the path and also works better with CI, because once run it will be cached under the user profile and it will not pull again.
More detail explanation at: https://tomgregory.com/what-is-the-gradle-wrapper-and-why-should-you-use-it/

kubernetes-kotlin/build.gradle.kts
```diff
@@ -0,0 +1,29 @@
...
+    maven { // The google mirror is less flaky than mavenCentral()
+        url = uri("https://maven-central.storage-download.googleapis.com/maven2/")
+    }
+    mavenCentral()
+    mavenLocal()
+}
+
+dependencies {
+    implementation("com.pulumi:pulumi:(,1.0]")
+    implementation("com.pulumi:kubernetes:3.19.1")
```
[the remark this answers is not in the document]
luistrigueiros: yes this work and its better actually

kubernetes-kotlin/src/main/kotlin/myproject/Support.kt
luistrigueiros: What version of the com.pulumi:kubernetes artifact is the new package available on ? 
I don't see it in com.pulumi:kubernetes:(,3.20]

luistrigueiros:
@t0yv0  yes I would like to also to make more Kotlin idiomatic but in a second iteration, first get as I need to read up a build of Type safe builders

kubernetes-kotlin/gradlew
stepango: Agree with @luistrigueiros including gradlew is considered the best practice

luistrigueiros:
I have update the template to use com.pulumi:kubernetes:(,4.0]

cnunciato:
Hi folks -- checking in on this one. @t0yv0 Would you be up for re-reviewing so we can bring it to a close, assuming it's something we feel like we should add?

t0yv0:
Sorry for the long delay here. Checking in again, it seems that it's currently blocked as it blows up with an exception due to a drift in pulumi-kubernetes falling behind on the java sdk-gen. I've opened an issue:

https://github.com/pulumi/pulumi-kubernetes/issues/2333

Issue #325, opened 2022-08-17
Remove .append files from Pulumi YAML templates


AaronFriel:
As of the following change, the .append file is no longer needed, and YAML templates must record all resources, variables, outputs, and configuration in the Pulumi.yaml file of the template.
- https://github.com/pulumi/pulumi/pull/10401

This change will be shipped in Pulumi 3.39, and by .40 or .41 we can remove the .append files from this repo.

cnunciato:
Hey @AaronFriel, checking in on this one. Is it safe to do this now?

PR #327, opened 2022-08-21
Add static-website templates

cnunciato:
This is a first pass at creating a set of static-website templates for AWS, Azure, and Google Cloud in TypeScript, Python, Go, C#, and YAML. Includes some canned web content, buckets configured for web access, and basic CDN configuration.

These can all be tested now by running:

```bash
pulumi new https://github.com/pulumi/templates/tree/cnunciato/static-website/{template-folder-name}
``` 

Fixes https://github.com/pulumi/devrel-team/issues/128.


static-website-aws-csharp/Program.cs
```diff
@@ -0,0 +1,35 @@
+using System.Collections.Generic;
+using Pulumi;
+using Aws = Pulumi.Aws;
+using SyncedFolder = Pulumi.SyncedFolder;
+
+return await Deployment.RunAsync(() => 
+{
+    var config = new Config();
+    var awsRegion = config.Get("awsRegion") ?? "us-west-2";
```
joeduffy, line 9: Should these be using the `aws:region` config?

static-website-aws-go/main.go
```diff
@@ -0,0 +1,54 @@
...
+		}
+		_, err = synced - folder.NewS3BucketFolder(ctx, "bucket-folder", &synced-folder.S3BucketFolderArgs{
+			Path:       pulumi.String(path),
+			BucketName: bucket.Bucket,
+			Acl:        pulumi.String("public-read"),
+		})
+		if err != nil {
+			return err
+		}
+		ctx.Export("url", bucket.WebsiteEndpoint.ApplyT(func(websiteEndpoint string) (string, error) {
```
joeduffy, line 49: Might be slightly nicer using the `pulumi.Sprintf` function?

static-website-aws-csharp/Program.cs
cnunciato: Indeed — sure what happened there, maybe something amiss with `pulumi convert`. Will take a look!

static-website-aws-go/main.go
cnunciato: Absolutely. These are all still the generated code from `pulumi convert’, but I plan to add some scripting to tidy things up.

aaronkao:
A piece of feedback while talking to Christian earlier

A distinction should be made between project config options, infrastructure config options, and the app level config options (e.g., index.html as default index document or the path)in the pulumi new project wizard in the CLI

static-website-aws-csharp/Program.cs
cnunciato: Fixed!

cnunciato:
All templates are tested and usable at this point, but the code for each one isn't quite where I'd like to be yet. (Still all generated from `pulumi convert` with a handful of fixes.) Will take care of that next.

cnunciato:
I believe the test failures here are unrelated to the content of the templates. @t0yv0 @stack72 is this something the platform team is still looking investigating?

static-website-aws-csharp/Program.cs
```diff
@@ -0,0 +1,114 @@
...
+    // Import the program's configuration settings.
+    var config = new Config();
+    var path = config.Get("path") ?? "./site";
+    var indexDocument = config.Get("indexDocument") ?? "index.html";
+    var errorDocument = config.Get("errorDocument") ?? "error.html";
+
+    // Create an S3 bucket and configure it as a website.
+    var bucket = new Aws.S3.Bucket("bucket", new()
+    {
+        Acl = "public-read",
```
cnunciato, line 17: I'll run through and add some enums as well for the languages that have them for these.

cnunciato: Actually on second thought, I'll do this as a follow-up, as they'll need to be added to Synced Folder as well, and I don't want to block on that. Easy enough to add post-release.

static-website-azure-csharp/Pulumi.yaml
```diff
@@ -0,0 +1,19 @@
+name: ${PROJECT}
+description: ${DESCRIPTION}
+runtime: dotnet
+
+template:
+  description: A static website on AWS
```
nimbinatus, line 6:
```suggestion
  description: A static website on Azure
```

static-website-azure-go/Pulumi.yaml
```diff
@@ -0,0 +1,19 @@
+name: ${PROJECT}
+description: ${DESCRIPTION}
+runtime: go
+
+template:
+  description: A static website on AWS
```
nimbinatus, line 6:
```suggestion
  description: A static website on Azure
```

static-website-azure-python/Pulumi.yaml
```diff
@@ -0,0 +1,19 @@
+name: ${PROJECT}
+description: ${DESCRIPTION}
+runtime: python
+
+template:
+  description: A static website on AWS
```
nimbinatus, line 6:
```suggestion
  description: A static website on Azure
```

static-website-azure-typescript/Pulumi.yaml
```diff
@@ -0,0 +1,19 @@
+name: ${PROJECT}
+description: ${DESCRIPTION}
+runtime: nodejs
+
+template:
+  description: A static website on AWS
```
nimbinatus, line 6:
```suggestion
  description: A static website on Azure
```

static-website-azure-yaml/Pulumi.yaml
```diff
@@ -0,0 +1,101 @@
+name: ${PROJECT}
+description: ${DESCRIPTION}
+runtime: yaml
+
+template:
+  description: A static website on AWS
```
nimbinatus, line 6:
```suggestion
  description: A static website on Azure
```

static-website-aws-python/__main__.py
```diff
@@ -0,0 +1,80 @@
...
+index_document = config.get("indexDocument") or "index.html"
+error_document = config.get("errorDocument") or "error.html"
+
+# Create an S3 bucket and configure it as a website.
+bucket = aws.s3.Bucket("bucket",
+    acl="public-read",
+    website=aws.s3.BucketWebsiteArgs(
+        index_document=index_document,
+        error_document=error_document,
+    ))
```
nimbinatus, line 17: Generally, PEP8 would have one of the following adjustments. It is *not* a blocker, and I'm not going to comment on every single one here because that's up to you:
All arguments at the same indent level, individual lines:
```suggestion
bucket = aws.s3.Bucket(
    "bucket",
    acl="public-read",
    website=aws.s3.BucketWebsiteArgs(
        index_document=index_document,
        error_document=error_document,
    ))
```
OR
All arguments aligned with the opening delimiter:
```suggestion
bucket = aws.s3.Bucket("bucket",
                       acl="public-read",
                       website=aws.s3.BucketWebsiteArgs(
                           index_document=index_document,
                           error_document=error_document,
                       ))
```

static-website-aws-python/pyvenv.cfg
```diff
@@ -0,0 +1,3 @@
+home = /usr/local/bin
```
nimbinatus, line 1: I wouldn't commit the `pyvenv.cfg` file. It generally isn't something we send in a repo (it lives in the `venv` directory if you're using the built-in virtual environment tool, which doesn't get committed), and someone will wonder why it appears. Pulumi creates the virtual environment based on the local Python environment, instead, and this file doesn't force it to use a specific version or anything.

static-website-azure-python/pyvenv.cfg
```diff
@@ -0,0 +1,3 @@
+home = /usr/local/bin
```
nimbinatus, line 1: See prior note re:removal

static-website-gcp-python/pyvenv.cfg
```diff
@@ -0,0 +1,3 @@
+home = /usr/local/bin
```
nimbinatus, line 1: See prior note re:removal

static-website-azure-python/__main__.py
```diff
@@ -0,0 +1,73 @@
+import pulumi
+import pulumi_azure_native as azure_native
+import pulumi_synced_folder as synced_folder
+import urllib
```
nimbinatus, line 4: Are you specifically trying to use only built-in libraries? I generally only find `urllib` versus `requests` or `urllib3` when people are trying to restrict their dependency/security surface area.

nimbinatus: Also, if you're only using one method or class, it's generally better practice to import just that need to avoid polluting your namespace, similar to the following:
```
from urllib.parse import urlparse
```

Finally, if you're going to import built-ins, generally PEP8 style has built-ins first (standard library modules), then third-party, then local (https://peps.python.org/pep-0008/#imports).

nimbinatus (changes requested):
There's a couple places in the YAML files that need to get updated. The other stuff is more nits or questions.

static-website-aws-python/pyvenv.cfg
cnunciato: Ah, good spot! I actually had a line in my code to delete these files, but it had a typo, and I didn't notice them sneaking through. Will remove. Thanks!

static-website-azure-python/__main__.py
cnunciato: >Are you specifically trying to use only built-in libraries?

Ideally (and I guess in general) yes; assuming this works and isn't somehow problematic, I figure folks can opt into a different library if they want to.

cnunciato: > PEP8 style has built-ins first (standard library modules), then third-party, then local

Ok thanks. So like this?

```python
from urllib.parse import urlparse
import pulumi
import pulumi_azure_native as azure_native
import pulumi_synced_folder as synced_folder
```

static-website-aws-python/__main__.py
cnunciato: Ok great, thanks. I think I'll go with your first suggestion, then, as that's also what I get out of the box when I run `black` against these files -- and I personally find that version a bit more readable as well. 😄

scottslowe:
@cnunciato In the `template` section, is there a `description` field supported for each key? I'm just wondering if it would be useful to have a prompt there to explain what each of the prompts is asking about.

static-website-azure-yaml/Pulumi.yaml
```diff
@@ -29,14 +29,6 @@ configuration:
     type: String
     default: error.html
 
-variables:
```
cnunciato: Committed by mistake — will fix.

static-website-azure-python/__main__.py
nimbinatus: > > Are you specifically trying to use only built-in libraries?
> 
> Ideally (and I guess in general) yes; assuming this works and isn't somehow problematic, I figure folks can opt into a different library if they want to.

Makes sense; it *is* a common practice if you're trying to reduce surface area. I probably would leave a comment in the line, then, just so newer Python folks aren't confused (most tutorials/self-guided learning systems use requests or urllib3; they very rarely use the built-in. If you want to explore why, check out https://realpython.com/urllib-request/#the-request-package-ecosystem, which has a very good writeup). 

If you add a comment, two spaces after the end of the import will make every linter happy :)

```python
import package  # Here's a comment
```

scottslowe (approved):

static-website-azure-python/__main__.py
nimbinatus: > > PEP8 style has built-ins first (standard library modules), then third-party, then local
> 
> Ok thanks. So like this?
> 
> ```python
> from urllib.parse import urlparse
> import pulumi
> import pulumi_azure_native as azure_native
> import pulumi_synced_folder as synced_folder
> ```

Yep, that technically meets PEP8. A lot of folks follow a slightly modified version that puts the `from .. import ..` calls at the bottom in the same order. Let me add a couple packages to show you what I mean:

```python
import os
import sys

import pulumi
import pulumi_azure_native as azure_native
import pulumi_synced_folder as synced_folder

from urllib.parse import urlparse

from requests.exceptions import HTTPError
```

**However**, that gets into arguments as to which style to follow that are more pedantic than helpful. As long as you follow PEP8 and the code is readable, you're good.

static-website-aws-python/__main__.py
nimbinatus: "Readability counts." 😉

cnunciato:
@pulumi/platform So it looks like the tests are all passing for the static-website templates (🎉) except for GCP, which is complaining about the number of GlobalAddress resources being hit:

https://github.com/pulumi/templates/runs/8116507581?check_suite_focus=true#step:19:3182

The template uses the [GlobalAddress](https://www.pulumi.com/registry/packages/gcp/api-docs/compute/globaladdress/) resource to enable CDN functionality for users out of the box. Is there some way we can suppress these particular tests somehow?

static-website-gcp-yaml/Pulumi.yaml
```diff
@@ -0,0 +1,86 @@
...
+
+resources:
+
+  bucket:
+    type: gcp:storage:Bucket
+    properties:
+      location: US
+      website:
+        mainPageSuffix: ${indexDocument}
+        notFoundPage: ${errorDocument}
```
AaronFriel, line 39: Nit: The next CLI release does not use the `.append` file, so any comments we want them to see should appear in both until we land #325.

AaronFriel (commented):
@cnunciato left a comment about .append files.

static-website-gcp-yaml/Pulumi.yaml
cnunciato: Ah great, ok thanks @AaronFriel. Will drop these in there as well, thn.

PR #328, opened 2022-08-22
Consistent test options

t0yv0:
Do not use Quick: True when performance testing, as it skews the benchmarks. 

https://github.com/pulumi/home/issues/1498

Test run: https://github.com/pulumi/templates/runs/7954302544?check_suite_focus=true

### DIFF
--- a/tests/template_test.go
+++ b/tests/template_test.go
@@ -62,7 +62,7 @@ func TestTemplates(t *testing.T) {
 
 	base := integration.ProgramTestOptions{
 		ExpectRefreshChanges:   true,
-		Quick:                  true,
+		Quick:                  !isTracingEnabled(),
 		SkipRefresh:            true,
 		NoParallel:             true, // we mark tests as Parallel manually when instantiating
 		DestroyOnCleanup:       true,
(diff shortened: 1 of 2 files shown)


tests/template_test.go
```diff
@@ -62,7 +62,7 @@ func TestTemplates(t *testing.T) {
 
 	base := integration.ProgramTestOptions{
 		ExpectRefreshChanges:   true,
-		Quick:                  true,
+		Quick:                  !isTracingEnabled(),
```
dixler, line 65: could you add the comment you put in the PR description?
> Do not use Quick: True when performance testing, as it skews the benchmarks.

tests/template_test.go
```diff
@@ -60,9 +60,14 @@ func TestTemplates(t *testing.T) {
 		templateUrl = specificTemplate
 	}
 
+	// When tracing is enabled to collect performance data, using
+	// Quick: true skews the measurements, therefore prefor Quick:
```
dixler, line 64: typo
```suggestion
	// Quick: true skews the measurements, therefore prefer Quick:
```

dixler (approved):
LGTM if tests pass

PR #329, opened 2022-08-23
Fix azure-fsharp template to use CreateSecret

Frassle:
The csharp template makes this a secret, the fsharp template should to.

### DIFF
--- a/azure-fsharp/Program.fs
+++ b/azure-fsharp/Program.fs
@@ -1,5 +1,6 @@
 ﻿module Program
 
+open Pulumi
 open Pulumi.FSharp
 open Pulumi.AzureNative.Resources
 open Pulumi.AzureNative.Storage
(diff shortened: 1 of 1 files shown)


azure-fsharp/Program.fs
```diff
@@ -21,7 +22,7 @@ let infra () =
     let primaryKey =
         ListStorageAccountKeysInvokeArgs(ResourceGroupName = resourceGroup.Name, AccountName = storageAccount.Name)
         |> ListStorageAccountKeys.Invoke
-        |> Outputs.apply (fun storageKeys -> storageKeys.Keys[0].Value)
+        |> Outputs.apply (fun storageKeys -> Output.CreateSecret(storageKeys.Keys[0].Value))
```
mikhailshilkov, line 25: I'm surprised this works - would expect you'd need to change apply to bind?

Frassle: hah yes I think this is double nesting the output like this

mikhailshilkov (approved):

PR #330, opened 2022-08-25
Fix make test_template targets

t0yv0:
I'm not entirely sure what the go test is doing by default to find the list of templates, but it's not scanning the folder. So when adding new templates make test_tempate target is broken. This works around it.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,7 +9,7 @@ test_templates::
 # Example: make test_template.typescript
 # This will run a test corresponding to the typescript template.
 test_template.%:
-	cd tests && BLACK_LISTED_TESTS=none go test -run "TestTemplate/^$*$$" $(TESTFLAGS)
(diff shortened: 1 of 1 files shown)


kpitzen (approved):

t0yv0:
@stack72 you had some comments in chat giving the background on why we have the current logic that selects only specific tests to run. But I wasn't sure if this change is unsafe in some way? It seems to me this is still a safe addition and it helps test things locally..

cnunciato:
This doesn't seem to cause my tests to run locally, but I'm not sure it's supposed to. Should we merge it now to see if it results in the tests being run in CI?

t0yv0:
This PR does not affect CI but it does unblock running tests locally before they are checked in (as form your branch) via:

```
make test_template.<template-name>
```

This was waiting on Paul Stack review though, I think this is not how we want to keep things. I'll just close for now.

PR #331, opened 2022-08-25
More generous timeouts

t0yv0:
https://github.com/pulumi/templates/actions/workflows/cron.yml seeing quite a few timeout related failures. It's not great we have templates timing out in 30min but some of it seems to be an artifact of the CI environment (cannot repro locally). The higher timeout would let us focus on the few tests that do get stuck.


aq17 (approved):
👍🏼

PR #336, opened 2022-08-26
Try caching NPM and pulumi plugins

t0yv0:
An experiment to add caching see if we can get any improvement on the timings of azure-javascript etc tests getting us:

- tighter timeouts on tests
- perhaps less network dependency-download-related interference on benchmarks for these


t0yv0:
This was a useful experiment but perhaps we do not want to do this, as it can interfere with result validity in case of fuzzy versions resolved from the cache hits.

PR #340, opened 2022-08-30
[WIP] Add templates for Kubernetes on AWS

scottslowe:
Add YAML template for standing up Kubernetes (EKS) on AWS in a new VPC

### DIFF
--- a/kubernetes-aws-yaml/Pulumi.yaml
+++ b/kubernetes-aws-yaml/Pulumi.yaml
@@ -1,0 +1,52 @@
+name: k8s-aws-yaml
+runtime: yaml
+description: A Pulumi YAML program to stand up a Kubernetes cluster on AWS
+template:
+  config:
+    aws:region:
+      description: The AWS region to deploy into
+      default: us-west-2
+configuration:
+  minClusterSize:
+    type: Number
+    default: 3
+  maxClusterSize:
+    type: Number
+    default: 6
+  desiredClusterSize:
+    type: Number
+    default: 3
+  eksNodeInstanceType:
+    type: String
+    default: t2.medium
+  vpcNetworkCidr:
+    type: String
+    default: 10.0.0.0/16
+resources:
+  # Create a VPC for the EKS cluster
+  eks-vpc:
+    type: awsx:ec2:Vpc
+    properties:
+      enableDnsHostnames: true
+      cidrBlock: ${vpcNetworkCidr}
+  # Create the EKS cluster
+  eks-cluster:
+    type: eks:Cluster
+    properties:
+      # Put the cluster in the new VPC created earlier
+      vpcId: ${eks-vpc.vpcId}
+      # Public subnets will be used for load balancers
+      publicSubnetIds: ${eks-vpc.publicSubnetIds}
+      # Private subnets will be used for cluster nodes
+      privateSubnetIds: ${eks-vpc.privateSubnetIds}
+      # Change configuration section above to change any of the following settings
+      instanceType: ${eksNodeInstanceType}
+      desiredCapacity: ${desiredClusterSize}
+      minSize: ${minClusterSize}
+      maxSize: ${maxClusterSize}
+      # Do not give the worker nodes public IP addresses
+      nodeAssociatePublicIpAddress: false
+outputs:
+  # Output the Kubeconfig for the cluster
+  kubeconfig: ${eks-cluster.kubeconfig}
+  vpcId: ${eks-vpc.vpcId}
--- a/kubernetes-aws-yaml/Pulumi.yaml.append
+++ b/kubernetes-aws-yaml/Pulumi.yaml.append
@@ -1,11 +1,3 @@
-name: ${PROJECT}
-runtime: yaml
-description: ${DESCRIPTION}
-template:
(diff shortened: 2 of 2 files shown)


rawkode (approved):

nimbinatus (commented):
The YAML file appears fine to me, but should there be a `Pulumi.yaml.append` file as in the other YAML templates?

scottslowe:
@nimbinatus I don't know. Let me look into this and I will respond back here when I have more information.

scottslowe:
It would appear that `Pulumi.yaml.append` isn't needed moving forward.

@lukehoban Can you have a look and let me know if this is what you had in mind, or if I should be incorporating/including something more in this base template?

nimbinatus (approved):

kubernetes-aws-yaml/Pulumi.yaml
```diff
@@ -0,0 +1,52 @@
+name: k8s-aws-yaml
+runtime: yaml
+description: A Pulumi YAML program to stand up a Kubernetes cluster on AWS
+template:
+  config:
+    aws:region:
+      description: The AWS region to deploy into
+      default: us-west-2
```
cnunciato, line 8: Replacing these with expressions will swap in these user-defined values as project creation-time:

```suggestion
name: ${PROJECT}
description: ${DESCRIPTION}
runtime: yaml

template:
  description: A Pulumi YAML program to stand up a Kubernetes cluster on AWS
  config:
    aws:region:
      description: The AWS region to deploy into
      default: us-west-2
```

kubernetes-aws-yaml/Pulumi.yaml
```diff
@@ -0,0 +1,52 @@
...
+configuration:
+  minClusterSize:
+    type: Number
+    default: 3
+  maxClusterSize:
+    type: Number
+    default: 6
+  desiredClusterSize:
+    type: Number
+    default: 3
```
cnunciato, line 18: When I run this, I get complaints about the default not being a string. I _think_ that means these should be:

 ```suggestion
  minClusterSize:
    type: Number
    default: "3"
  maxClusterSize:
    type: Number
    default: "6"
  desiredClusterSize:
    type: Number
    default: "3"
```

**Review by cnunciato — APPROVED** [review]
Couple little things, but otherwise 👍! 🎉 

It's perhaps worth mentioning that when I went to run this, I got an error about a missing plugin:

```
$ pulumi up
...

error: Error: error resolving type of resource eks-vpc: 
    internal error loading package "awsx": 
        no resource plugin 'pulumi-resource-awsx' found in the workspace or on your $PATH, install the plugin using `pulumi plugin install resource awsx`
```

... and the remedy supplied by the CLI didn't work:

```
$ pulumi plugin install resource awsx
[resource plugin awsx-0.40.0] installing
warning: Error downloading plugin: 403 HTTP error fetching plugin from https://get.pulumi.com/releases/plugins/pulumi-resource-awsx-v0.40.0-darwin-amd64.tar.gz
```

Installing the beta got me past it, though:

```
pulumi plugin install resource awsx v1.0.0-beta.7
```

I don't think there's anything we can do at the template level for this, of course -- I just wanted to point it out, as there's a good chance even with the fix going in with https://github.com/pulumi/pulumi/pull/10530 that the CLI will fail to recognize that it needs to install the beta. So if that's the case, we may just need to remember to include a note in the docs that this step may need to be done manually for now.

File: kubernetes-aws-yaml/Pulumi.yaml [hunk 3]
```diff
@@ -0,0 +1,52 @@
+name: k8s-aws-yaml
```
cnunciato, line 1: I know we have a fix going out soon to stop requiring `.append` files (yay), but for now I think we still need them. (Otherwise when run with `pulumi new`, we'll create a stack with no resources.)

kubernetes-aws-yaml/Pulumi.yaml
scottslowe: @cnunciato I can't reproduce this on my system. When I run `pulumi new <path-to-template>` and then run `pulumi up` from the resulting new project, I don't get any messages or complaints about the defaults.

kubernetes-aws-yaml/Pulumi.yaml
scottslowe: Changes incorporated via e084496

kubernetes-aws-yaml/Pulumi.yaml
scottslowe: Resolved via e084496 (added `Pulumi.yaml.append`)

aaronkao:
Let's generate the other languages first before merging the PR

scottslowe:
@aaronkao That's fine; I just wanted "approval" on the YAML before generating the other languages. That way, if changes are needed I can reflect those in the YAML _once_ instead of having to make multiple identical (or nearly-identical) changes in each of the languages.

scottslowe:
I'm going to close this PR in favor of a new one that has working versions of the template in Go, Typescript, Python, and YAML (Code for Java and C# will be generated but may not work.) The new PR will also try to better adhere to naming standards and conventions.

PR #362, opened 2022-09-14
Apply prewarm when benchmarking

t0yv0:
Wanted to make a couple of changes to improve benchmarking cron. As part of this, realized that the use cases for the test cron and benchmarking cron are diverging so I factored them out into separate code paths for easier reading.

Changes:

- prewarm phase to rule out any further/late plugin installation
- separate SKIP config, skip static-website templates for now
- use dev versions of Pulumi CLI instead of release versions; faster feedback
    - had to roll this back since with bors changes in p/p we do not have a usable "dev" pointer at the moment  
- remove parallelism from benchmarks to get better numbers
- higher duration of aws token session to compensate for longer runtimes and avoid failures to upload data

Also I've gotten to the bottom of PULUMI_TEMPLATE_LOCATION use I think - if unset, this tests released templates, but if set to PWD it uses locally modified copies. This is setup just fine on CI but is confusing locally when trying to test out template changes on my machine, by default the tests just ignore those changes unless I remember to set this env var. I've setup convenient make targets and documented this detail, but there's no changes to how CI runs things with respect to this env var.

### DIFF
--- a/.github/workflows/performance_metrics_cron.yml
+++ b/.github/workflows/performance_metrics_cron.yml
@@ -72,7 +72,7 @@ jobs:
           aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
           aws-region: ${{ env.AWS_REGION }}
           aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
-          role-duration-seconds: 14400
+          role-duration-seconds: 14400 # 4 hours
           role-session-name: templates@githubActions
           role-to-assume: ${{ secrets.AWS_CI_ROLE_ARN }}
       - uses: actions/checkout@v2
--- a/.github/workflows/performance_metrics_cron.yml
+++ b/.github/workflows/performance_metrics_cron.yml
@@ -11,7 +11,7 @@ env:
   AWS_REGION: us-west-2
   GOOGLE_CREDENTIALS: ${{ secrets.GCP_CREDENTIALS }}
   LINODE_TOKEN: ${{ secrets.LINODE_TOKEN }}
-  SKIPPED_TESTS: "alicloud,digitalocean,kubernetes,openstack,equinix-metal,civo,aiven,auth0,github,yaml,oci,java-jbang,java-gradle,azuredevops"
+  SKIPPED_BENCHMARKS: "alicloud,digitalocean,kubernetes,openstack,equinix-metal,civo,aiven,auth0,github,yaml,oci,java-jbang,java-gradle,azuredevops,static-website"
   PULUMI_API: https://api.pulumi-staging.io
   SLACK_WEBHOOK_URL: ${{ secrets.SLACK_WEBHOOK_URL }}
   ARM_CLIENT_ID: ${{ secrets.ARM_CLIENT_ID }}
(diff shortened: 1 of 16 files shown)


t0yv0:
Test run https://github.com/pulumi/templates/actions/runs/3053618633

t0yv0:
Prewarm + no parallelism makes it take a lot more time but I feel it is worth it. The numbers popping out are smaller again, and hopefully the variance will be smaller also, making them much more useful to us.

.github/workflows/performance_metrics_cron.yml
```diff
@@ -85,12 +72,16 @@ jobs:
           aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
           aws-region: ${{ env.AWS_REGION }}
           aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
-          role-duration-seconds: 3600
+          role-duration-seconds: 14400
```
RobbieMcKinstry, line 75: I like to add inline unit conversion but it's easily inspectable without the comment too.
```suggestion
          role-duration-seconds: 14400 # 4 hours
```

.github/workflows/performance_metrics_cron.yml
```diff
@@ -109,10 +100,9 @@ jobs:
         continue-on-error: true
         run: |
           set -euo pipefail
-          cd tests && go test -v -json -count=1 -cover -timeout 6h -parallel ${{ env.TESTPARALLELISM }} . 2>&1 | tee /tmp/gotest.log | gotestfmt
+          cd tests/perf && go test -v -json -count=1 -timeout 6h . 2>&1 | tee /tmp/gotest.log | gotestfmt
```
RobbieMcKinstry, line 103: My understanding was without specifying parallelism, Go uses MAX_PARALLELISM=${cpu_count} by default (or 2*cpu_count). Hard to find documentation of this but [this](https://pkg.go.dev/testing#B.RunParallel) is as close as I can get.

RobbieMcKinstry, line 103:
```suggestion
          cd tests/perf && go test -v -json -count=1 -timeout 6h -parallel 1 . 2>&1 | tee /tmp/gotest.log | gotestfmt
```

Makefile
```diff
@@ -2,22 +2,65 @@ PROJECT_NAME 	:= Pulumi Templates
...
+#
+# This will run a typescript template test and populate `./traces`
+# with performance data. See also
+# https://www.pulumi.com/docs/support/troubleshooting/#performance
+#
+# Accepts the same environment variables as `test_template.%` target.
 bench_template.%:
-	mkdir -p ./traces
-	cd tests && PULUMI_TRACING_DIR=${PWD}/traces BLACK_LISTED_TESTS=none go test -run "TestTemplate/^$*$$" $(TESTFLAGS)
+	rm -rf ${PWD}/traces
```
RobbieMcKinstry, line 54: Running `rm -rf $ANYTHING/foobar` always scares me (in case e.g. $ANYTHING isnt the empty string, or doesn't point to where we think it does) but I can't think of a scenario where it goes wrong. Only if you were running `make -f` or something. It's probably ok to leave as is?

RobbieMcKinstry (approved):
Nice! I left a few comments at your discretion.

t0yv0:
@lblackstone adding in case you can represent providers team here - editing shared code, wanted to make sure it's ok.

lblackstone (approved):
LGTM

PR #364, opened 2022-09-15
Add templates for Kubernetes on AWS

scottslowe:
This PR adds templates for Kubernetes on AWS in YAML, Typescript, Python, and Go (C# and Java to be added later).


kubernetes-aws-python/__main__.py
```diff
@@ -0,0 +1,34 @@
...
+    vpc_id=eks_vpc.vpc_id,
+    public_subnet_ids=eks_vpc.public_subnet_ids,
+    private_subnet_ids=eks_vpc.private_subnet_ids,
+    instance_type=eks_node_instance_type,
+    desired_capacity=desired_cluster_size,
+    min_size=min_cluster_size,
+    max_size=max_cluster_size,
+    node_associate_public_ip_address=False)
+pulumi.export("kubeconfig", eks_cluster.kubeconfig)
+pulumi.export("vpcId", eks_vpc.vpc_id)
```
nimbinatus, line 34: Here's how I'd do it idiomatically.

```suggestion
import pulumi
import pulumi_awsx as awsx
import pulumi_eks as eks

config = pulumi.Config()

min_cluster_size = 3 if config.get_float("minClusterSize") is None else config.get_float("minClusterSize")
max_cluster_size = 6 if config.get_float("maxClusterSize") is None else config.get_float("maxClusterSize")
desired_cluster_size = 3 if config.get_float("desiredClusterSize") is None else config.get_float("desiredClusterSize")
eks_node_instance_type = "t2.Medium" if config.get("eksNodeInstanceType") is None else config.get("eksNodeInstanceType")
vpc_network_cidr = "10.0.0.0/16" if config.get("vpcNetworkCidr") is None else config.get("vpcNetworkCidr")

eks_vpc = awsx.ec2.Vpc(
    "eks-vpc",
    enable_dns_hostnames=True,
    cidr_block=vpc_network_cidr
)

eks_cluster = eks.Cluster(
    "eks-cluster",
    vpc_id=eks_vpc.vpc_id,
    public_subnet_ids=eks_vpc.public_subnet_ids,
    private_subnet_ids=eks_vpc.private_subnet_ids,
    instance_type=eks_node_instance_type,
    desired_capacity=desired_cluster_size,
    min_size=min_cluster_size,
    max_size=max_cluster_size,
    node_associate_public_ip_address=False
)

pulumi.export("kubeconfig", eks_cluster.kubeconfig)
pulumi.export("vpcId", eks_vpc.vpc_id)
```

nimbinatus: Also, I notice comments in the YAML version and the Go version, but not the other language versions. Would they be helpful?

kubernetes-aws-typescript/Pulumi.yaml
```diff
@@ -0,0 +1,10 @@
+name: ${PROJECT}
+description: ${DESCRIPTION}
+runtime: nodejs
+template:
+  description: A Nodejs program to create a Kubernetes cluster on AWS
```
nimbinatus, line 5: Would this be instead `A TypeScript program...`?

nimbinatus (commented):
I'm mainly reviewing the Python one for idiomatic use; I glanced through the others, but other folks probably are better at spotting faithful code style there than me.

kubernetes-aws-python/__main__.py
scottslowe: Yes, comments would be useful, I think. I'll add them.

To be sure I'm interpreting your idiomatic code correctly, is this what you're suggesting?
```
min_cluster_size = 3 if config.get_float("minClusterSize") is None else config.get_float("minClusterSize")
```
(all on one line, no line breaks)

kubernetes-aws-typescript/Pulumi.yaml
scottslowe: Fixed

scottslowe: Fixed in 73aaab4

kubernetes-aws-python/__main__.py
scottslowe: Changed in 73aaab4

nimbinatus: > To be sure I'm interpreting your idiomatic code correctly, is this what you're suggesting?
> ```
> min_cluster_size = 3 if config.get_float("minClusterSize") is None else config.get_float("minClusterSize")
> ```
> (all on one line, no line breaks)

Yes, that's the main change to make it more idiomatic. I'm also adding some line breaks and reformatting the instance creation to make it a bit cleaner and match a few different Python style guides.

kubernetes-aws-go/main.go
```diff
@@ -0,0 +1,72 @@
...
+			// Change configuration values above to change any of the following settings
+			InstanceType:                 pulumi.String(eksNodeInstanceType),
+			DesiredCapacity:              pulumi.Int(desiredClusterSize),
+			MinSize:                      pulumi.Int(minClusterSize),
+			MaxSize:                      pulumi.Int(maxClusterSize),
+			// Do not give the worker nodes a public IP address
+			NodeAssociatePublicIpAddress: pulumi.Bool(false),
+			// Uncomment the next two lines for a private cluster (VPN access required)
+			// EndpointPrivateAccess: 	      pulumi.Bool(true),
+			// EndpointPublicAccess:         pulumi.Bool(false),
```
cnunciato, line 61: I haven't spent much time with Kubernetes, so since I don't have a sense of common it is for someone to think about this (i.e., whether to make the cluster private or not) -- but if it's a common thing, I wonder if exposing it as an optional config setting, false by default like you have here, and then using that setting to set these values accordingly in all cases would be better than comments. That way, we'd be able to validate these properties as well, also and also be able to test both code paths if we wanted. 

Alternatively, we could leave these lines out, and handle the use case as a next-steps item in the template pages. That might give it a bit more room for explanation as well. Just some thoughts, wonder what you think.

kubernetes-aws-go/Pulumi.yaml
```diff
@@ -0,0 +1,10 @@
+name: ${PROJECT}
+description: ${DESCRIPTION}
+runtime: go
+template:
+  description: A Go program to create a Kubernetes cluster on AWS
+  important: true
```
cnunciato, line 6: I'm actually not sure whether we should or should not be setting this for these arch templates. I didn't set it for the static-website templates, and it doesn't look like Luke's PR sets it, either. This used to toggle whether the template appeared in the list of options with `pulumi new` called with no arguments, but the CLI I have today doesn't seem to use this value anymore. (I [asked about this today](https://pulumi.slack.com/archives/C9SGX9QA1/p1663283005951889) and it sounds like this might be a bug, but at the same time, the UX around this comment is also changing, so we may or may not use this property going forward.) 

If it were me, I'd probably leave it out for now, just for consistency, but yeah -- we'll know more soon. 😄

kubernetes-aws-typescript/package.json
```diff
@@ -0,0 +1,12 @@
...
+		"devDependencies": {
+			"@types/node": "^14"
+		},
+		"dependencies": {
+			"typescript": "^4.0.0",
+			"@pulumi/pulumi": "^3.0.0",
+			"@pulumi/awsx": "*",
+			"@pulumi/eks": "*"
+		}
+}
```
cnunciato, line 12: I know it's a codegen thing, but we should probably tidy these up to be 4 spaces. (My bad for not filing an issue on this, as I noticed it also -- I'll do that.)

kubernetes-aws-yaml/Pulumi.yaml
```diff
@@ -0,0 +1,57 @@
...
+    default: 6
+  desiredClusterSize:
+    type: Number
+    default: 3
+  eksNodeInstanceType:
+    type: String
+    default: t2.medium
+  vpcNetworkCidr:
+    type: String
+    default: 10.0.0.0/16
```
cnunciato, line 26: Would it make sense to prompt for any of these (or some subset of them) during `pulumi new`? Seems like that might be nice, but I defer to you on it. If you think so too, we should duplicate them in under `template.config` (and add some descriptions).

cnunciato (commented):
This looks great! Bit of initial feedback for you to consider. Are you planning on adding C# in this PR, or in a follow-up?

kubernetes-aws-python/__main__.py
scottslowe: @nimbinatus Your comment should be addressed, then; let me know if it isn't!

kubernetes-aws-python/__main__.py
```diff
@@ -0,0 +1,40 @@
...
+import pulumi_awsx as awsx
+import pulumi_eks as eks
+
+# Get some values from the Pulumi configuration (or use defaults)
+config = pulumi.Config()
+min_cluster_size = 3 if config.get_float("minClusterSize") is None else config.get_float("minClusterSize")
+max_cluster_size = 6 if config.get_float("maxClusterSize") is None else config.get_float("maxClusterSize")
+desired_cluster_size = 3 if config.get_float("desiredClusterSize") is None else config.get_float("desiredClusterSize")
+eks_node_instance_type = "t2.medium" if config.get("eksNodeInstanceType") is None else config.get("eksNodeInstanceType")
+vpc_network_cidr = "10.0.0.0/16" if config.get("vpcNetworkCidr") is None else config.get("vpcNetworkCidr")
```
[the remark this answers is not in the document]
scottslowe: Haha, it was @nimbinatus that suggested the current syntax! 🤣

kubernetes-aws-go/Pulumi.yaml
scottslowe: @cnunciato Are you referring to the `important: true` line?

kubernetes-aws-go/main.go
scottslowe: The "best practices" configuration would make the cluster private, but since _some_ sort of connectivity into the VPC (VPN or similar) would be required I opted against a private cluster and chose again to expose the option with a comment. I did (sort of) validate these properties by running a `pulumi preview` with them uncommented, but your suggestion to change the values and expose them uncommented may be worthwhile.

kubernetes-aws-python/__main__.py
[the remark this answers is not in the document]
nimbinatus: ... I just realized that the get methods for Pulumi config are actually using the dictionary get methods 🤦‍♀️  Sorry! The version I was doing was for general default values (it handles falsey values differently than a simple or). I thought they had made new get methods instead of using the underlying method, which actually can handle default values natively. https://www.pulumi.com/docs/reference/pkg/python/pulumi/#pulumi.Config.get_float New syntax:

```python
min_cluster_size = config.get_float("minClusterSize", 3)
max_cluster_size = config.get_float("maxClusterSize", 6)
desired_cluster_size = config.get_float("desiredClusterSize", 3)
eks_node_instance_type = config.get("eksNodeInstanceType", "t2.medium")
vpc_network_cidr = config.get("vpcNetworkCidr", "10.0.0.0/16")
```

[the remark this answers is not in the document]
cnunciato: Oh, nice! I didn't know this worked for Python. Much nicer. Thanks @nimbinatus!

kubernetes-aws-go/Pulumi.yaml
cnunciato: Yep, sorry -- I see now I could've made that clearer. 😂

kubernetes-aws-python/__main__.py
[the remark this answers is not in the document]
scottslowe: Fixed in 8f2808f

kubernetes-aws-typescript/package.json
scottslowe: Fixed. Incorrect version reference for `@pulumi/awsx` addressed as well. Changes are in 8f2808f.

kubernetes-aws-yaml/Pulumi.yaml
scottslowe: Fixed in 8f2808f. All languages (YAML, Go, Python, TypeScript) now prompt the user for these values when using `pulumi new`.

kubernetes-aws-go/Pulumi.yaml
scottslowe: Fixed in 8f2808f. The `important: true` line is removed from `Pulumi.yaml` for all languages present in this PR.

kubernetes-aws-typescript/Pulumi.yaml
```diff
@@ -0,0 +1,10 @@
+name: ${PROJECT}
+description: ${DESCRIPTION}
+runtime: nodejs
+template:
+  description: A Typescript program to create a Kubernetes cluster on AWS
```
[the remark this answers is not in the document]
scottslowe: Fixed in 8f2808f

scottslowe:
> This looks great! Bit of initial feedback for you to consider. Are you planning on adding C# in this PR, or in a follow-up?

@cnunciato I plan to add C# in a follow-up PR (and then Java later when it's ready, of course).

nimbinatus (commented):
Python code LGTM

cnunciato (approved):
Looks good! Nothing need block. I'd still love to see those commented lines be config-driven! But we can come back to that later. Awesome work! 🎉

cnunciato:
@scottslowe These test failures are not related to your PR, so I'd go ahead and merge them.

PR #375, opened 2022-09-25
Add Kubernetes on Azure templates

scottslowe:
This PR adds templates for Kubernetes on Azure in Go, Python, TypeScript, and YAML. C# and Java will come in follow-up PRs.


kubernetes-azure-python/__main__.py
```diff
@@ -0,0 +1,120 @@
...
+
+import base64
+import pulumi
+from pulumi_azure_native import resources
+from pulumi_azure_native import network
+from pulumi_azure_native import containerservice
+
+# Get some values from the Pulumi configuration (or use defaults)
+azure_cfg = pulumi.Config("azure-native")
+azureLocation = azure_cfg.require("location")
```
nimbinatus, line 11: any specific reason you're flipping between `require()` and `get()` here?

kubernetes-azure-python/__main__.py
```diff
@@ -0,0 +1,120 @@
...
+proj_cfg = pulumi.Config()
```
nimbinatus, line 12: Any reason to have a duplicate assignment here with lines 10 and 12? I'm guessing this is a namespacing thing, but I think you probably want a comment explaining that so people don't immediately remove this line and change the `proj_cfg` to `azure_cfg` and break things.

kubernetes-azure-python/__main__.py
```diff
@@ -0,0 +1,120 @@
...
+num_worker_nodes = proj_cfg.get_float("numWorkerNodes", 3)
+k8s_version = proj_cfg.get("kubernetesVersion", "1.24.3")
+prefix_for_dns = proj_cfg.get("prefixForDns", "pulumi")
+node_vm_size = proj_cfg.get("nodeVmSize", "Standard_DS2_v2")
+mgmt_group_id = proj_cfg.require("mgmtGroupId")
+ssh_pub_key = proj_cfg.require("sshPubKey")
+
+# Create an Azure Resource Group
+resource_group = resources.ResourceGroup("resource_group",
```
nimbinatus, line 21: [PEP8](https://peps.python.org/pep-0008/#indentation) linters will flag all of these instances as not being aligned to PEP8. You have two options:
**Option A**:
```python
resource_group = resources.ResourceGroup("resource_group",
                                         location=azureLocation,
                                         resource_group_name="aks-rg"
                                         )
```

**Option B** (my personal preference because it reduces the line-too-long errors that will happen from other linters):
```python
resource_group = resources.ResourceGroup(
    "resource_group",
    location=azureLocation,
    resource_group_name="aks-rg"
)
```

kubernetes-azure-python/__main__.py
```diff
@@ -0,0 +1,120 @@
...
+    location=azureLocation,
+    resource_group_name="aks-rg"
+    )
+
+# Create an Azure Virtual Network
+virtual_network = network.VirtualNetwork("virtual_network",
+    address_space=network.AddressSpaceArgs(
+        address_prefixes=["10.0.0.0/16"],
+        ),
```
nimbinatus, line 30:
```suggestion
    address_space=network.AddressSpaceArgs(
        address_prefixes=["10.0.0.0/16"],
    ),
```

nimbinatus (commented):
A couple notes on the Python part; I haven't reviewed the others.

kubernetes-azure-python/__main__.py
scottslowe: Yes. The `azure-native:location` configuration value _must_ be present (required by the provider as well as being used in the code), whereas the others have default values supplied if the configuration value isn't present. Is there a better/more recommended way of handling this sort of situation?

kubernetes-azure-python/__main__.py
scottslowe: You are correct that it is due to configuration namespacing. I'll add a comment to that effect (and in other languages where a similar mechanism is needed).

scottslowe: Resolved via a comment added in c4dea65

kubernetes-azure-python/__main__.py
scottslowe: Switched to Option B in c4dea65, should be resolved now

kubernetes-azure-python/__main__.py
scottslowe: Fixed in c4dea65

kubernetes-azure-python/__main__.py
nimbinatus: Hm. I honestly don't know; what you're saying makes sense, but I've never used both in the same program, and I don't think the SDK docs make that clear. Maybe adjust the comment on line 9 to something like this?
```python
# Either pull required values from the Pulumi configuration, or get values and use defaults if null
```

kubernetes-azure-python/__main__.py
nimbinatus: Hm, not quite right. I'll make suggestions.

kubernetes-azure-python/__main__.py
```diff
@@ -0,0 +1,122 @@
...
+prefix_for_dns = proj_cfg.get("prefixForDns", "pulumi")
+node_vm_size = proj_cfg.get("nodeVmSize", "Standard_DS2_v2")
+mgmt_group_id = proj_cfg.require("mgmtGroupId")
+ssh_pub_key = proj_cfg.require("sshPubKey")
+
+# Create an Azure Resource Group
+resource_group = resources.ResourceGroup("resource_group",
+    location=azureLocation,
+    resource_group_name="aks-rg"
+)
```
nimbinatus, line 26:
```suggestion
resource_group = resources.ResourceGroup(
    "resource_group",
    location=azureLocation,
    resource_group_name="aks-rg"
)
```

kubernetes-azure-python/__main__.py
```diff
@@ -0,0 +1,122 @@
...
+    virtual_network_name="aks-vnet"
+)
+
+# Create three subnets in the virtual network
+subnet1 = network.Subnet("subnet-1",
+    address_prefix="10.0.0.0/22",
+    resource_group_name=resource_group.name,
+    subnet_name="aks-subnet-1",
+    virtual_network_name=virtual_network.name
+)
```
nimbinatus, line 44:
```suggestion
subnet1 = network.Subnet(
    "subnet-1",
    address_prefix="10.0.0.0/22",
    resource_group_name=resource_group.name,
    subnet_name="aks-subnet-1",
    virtual_network_name=virtual_network.name
)
```

kubernetes-azure-python/__main__.py
```diff
@@ -0,0 +1,122 @@
...
+subnet2 = network.Subnet("subnet-2",
+    address_prefix="10.0.4.0/22",
+    resource_group_name=resource_group.name,
+    subnet_name="aks-subnet-2",
+    virtual_network_name=virtual_network.name
+)
```
nimbinatus, line 50:
```suggestion
subnet2 = network.Subnet(
    "subnet-2",
    address_prefix="10.0.4.0/22",
    resource_group_name=resource_group.name,
    subnet_name="aks-subnet-2",
    virtual_network_name=virtual_network.name
)
```

kubernetes-azure-python/__main__.py
```diff
@@ -0,0 +1,122 @@
...
+subnet3 = network.Subnet("subnet-3",
+    address_prefix="10.0.8.0/22",
+    resource_group_name=resource_group.name,
+    subnet_name="aks-subnet-3",
+    virtual_network_name=virtual_network.name
+)
```
nimbinatus, line 56:
```suggestion
subnet3 = network.Subnet(
    "subnet-3",
    address_prefix="10.0.8.0/22",
    resource_group_name=resource_group.name,
    subnet_name="aks-subnet-3",
    virtual_network_name=virtual_network.name
)
```

kubernetes-azure-python/__main__.py
```diff
@@ -0,0 +1,122 @@
...
+
+# Create an Azure Kubernetes Service cluster
+managed_cluster = containerservice.ManagedCluster("managed_cluster",
```
nimbinatus, line 59:
```suggestion
managed_cluster = containerservice.ManagedCluster(
    "managed_cluster",
```

nimbinatus (commented):
Like this for PEP8 for the Python code

kubernetes-azure-python/__main__.py
scottslowe: I'd overlooked another reason for using both `get()` and `require()` here. The last two project-namespaced configuration values (`mgmtGroupId` and `sshPubKey`) are values where I _can't_ provide a default value; they are too dependent on each customer's environment. I'll update the comments and add another one to better explain.

scottslowe: Comments re-worded and a new comment added in 5b8b6ce

kubernetes-azure-python/__main__.py
scottslowe: Should be fixed/correct in 5b8b6ce

kubernetes-azure-python/__main__.py
```diff
@@ -0,0 +1,122 @@
...
+
+# Create an Azure Virtual Network
+virtual_network = network.VirtualNetwork("virtual_network",
+    address_space=network.AddressSpaceArgs(
+        address_prefixes=["10.0.0.0/16"],
+    ),
+    location=azureLocation,
+    resource_group_name=resource_group.name,
+    virtual_network_name="aks-vnet"
+)
```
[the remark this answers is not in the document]
scottslowe: Should be fixed/corrected in 5b8b6ce

kubernetes-azure-python/__main__.py
scottslowe: Addressed in 5b8b6ce (should be correct now)

kubernetes-azure-python/__main__.py
scottslowe: Fixed in 5b8b6ce

kubernetes-azure-python/__main__.py
scottslowe: This change is included in 5b8b6ce

kubernetes-azure-python/__main__.py
cnunciato: This works, but since we panic when these two required values aren't provided (and there's no indication they're necessary until you run the `pulumi up` after the new-project wizard finishes), do you think it might be better to signal their requirement with config prompts, and just not supply default values?

scottslowe: We can certainly add them to the `template` portion of `Pulumi.yaml` and not supply a default value. For the management group, I think that would probably work fine. However, I suspect the SSH public key will present a problem. You can't use `pulumi config set sshPubKey <key-material-here>` as it will throw an error. You have to use `pulumi config set sshPubKey < <path-to-public-key>`. (It has something to do with multi-line values, AIUI.) Let me try adding them to the `template` portion and see if there are errors with the SSH public key.

kubernetes-azure-go/main.go
```diff
@@ -0,0 +1,183 @@
...
+		// Create an Azure Virtual Network
+		virtualNetwork, err := network.NewVirtualNetwork(ctx, "aks-network", &network.VirtualNetworkArgs{
+			AddressSpace: &network.AddressSpaceArgs{
+				AddressPrefixes: pulumi.StringArray{
+					pulumi.String("10.0.0.0/16"),
+				},
+			},
+			Location:           pulumi.String(azureLocation),
+			ResourceGroupName:  resourceGroup.Name,
+			VirtualNetworkName: pulumi.String("aks-network"),
```
cnunciato, line 56: Similar question here about hard-coding the name.

kubernetes-azure-python/__main__.py
cnunciato: FWIW, I've been using `black` to format my Python code and it's been working well (I think -- @nimbinatus will let me know if I'm wrong about that!).

kubernetes-azure-go/main.go
```diff
@@ -0,0 +1,183 @@
...
+		}
+		nodeVmSize, err := cfg.Try("nodeVmSize")
+		if err != nil {
+			nodeVmSize = "Standard_DS2_v2"
+		}
+
+		// Create an Azure Resource Group
+		resourceGroup, err := resources.NewResourceGroup(ctx, "resourceGroup", &resources.ResourceGroupArgs{
+			Location:          pulumi.String(azureLocation),
+			ResourceGroupName: pulumi.String("aks-rg"),
```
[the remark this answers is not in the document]
scottslowe: The `azureLocation` variable _is_ pulled from the provider configuration (see line 17).

As for the names, we can certainly remove those lines and let them be auto-generated. I left them there in the event the user wanted to specify a name, but perhaps letting them get auto-generated would be a better approach.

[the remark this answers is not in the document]
cnunciato: > The azureLocation variable is pulled from the provider configuration (see line 17).

Oh yes -- what I mean is that I _think_ if you _removed_ both line 17 and the references to `Location` throughout, things would "just work", since all Azure Native resources should (by default) use whatever location is configured _on_ the provider.

> As for the names, we can certainly remove those lines and let them be auto-generated.

Ok cool, yeah I'd definitely recommend that then.

kubernetes-azure-go/main.go
```diff
@@ -0,0 +1,183 @@
...
+		})
+		if err != nil {
+			return err
+		}
+		ctx.Export("resourceGroupName", resourceGroup.Name)
```
cnunciato, line 45: This is mostly anecdotal, but I think I've generally seen these clustered up at the end of the program, rather than sprinkled throughout, and this is generally how I've written my programs as well. Do you think we should be consistent there, or do you prefer this approach to that one?

kubernetes-azure-go/main.go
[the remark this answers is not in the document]
scottslowe: Fixed via 270555d

kubernetes-azure-go/main.go
scottslowe: Fixed via 270555d

kubernetes-azure-go/main.go
scottslowe: Fixed via 270555d

kubernetes-azure-python/__main__.py
scottslowe: I think all these Python formatting errors should be corrected now, thanks!

kubernetes-azure-python/__main__.py
scottslowe: Fixed in 270555d

scottslowe:
@cnunciato I've incorporated your feedback on the user experience (adding management group ID and SSH public key to the `template` portion) and I've removed unnecessary location and name settings.

@nimbinatus I believe I've addressed all the formatting issues.

This is ready for another review!

kubernetes-azure-go/Pulumi.yaml
```diff
@@ -0,0 +1,25 @@
...
+    kubernetesVersion:
+      default: 1.24.3
+      description: Kubernetes version to deploy in the cluster
+    nodeVmSize:
+      default: Standard_DS2_v2
+      description: VM size to use for worker nodes in the cluster
+    mgmtGroupId:
+      description: The object ID (from az ad group list) of a group to make cluster admins
+    sshPubKey:
+      description: The public key for SSH access to cluster nodes
```
cnunciato, line 25: Maybe a nit, but if I were reading this prompt as a first-time user, my one question would be, "Is this asking me for the content of the key itself or the path to the key file?"

kubernetes-azure-go/main.go
```diff
@@ -0,0 +1,171 @@
...
+		if err != nil {
+			numWorkerNodes = 3
+		}
+		nodeVmSize, err := cfg.Try("nodeVmSize")
+		if err != nil {
+			nodeVmSize = "Standard_DS2_v2"
+		}
+
+		// Create an Azure Resource Group
+		resourceGroup, err := resources.NewResourceGroup(ctx, "resourceGroup")
```
cnunciato, line 37:
```suggestion
		resourceGroup, err := resources.NewResourceGroup(ctx, "resourceGroup", &resources.ResourceGroupArgs{})
```

kubernetes-azure-go/main.go
```diff
@@ -0,0 +1,171 @@
...
+		subnet1, err := network.NewSubnet(ctx, "subnet-1", &network.SubnetArgs{
+			AddressPrefix:      pulumi.String("10.0.0.0/22"),
+			ResourceGroupName:  resourceGroup.Name,
+			VirtualNetworkName: virtualNetwork.Name,
+		})
+		if err != nil {
+			return err
+		}
+
+		subnet2, err := network.NewSubnet(ctx, "subnet-2", &network.SubnetArgs{
```
cnunciato, line 65: When I run this, I get compile errors about this being declared but not used. If we don't need to reference `subnet2` elsewhere in the code, we can ignore it with `_`:

```suggestion
		_, err = network.NewSubnet(ctx, "subnet-2", &network.SubnetArgs{
```

kubernetes-azure-go/main.go
```diff
@@ -0,0 +1,171 @@
...
+			AddressPrefix:      pulumi.String("10.0.4.0/22"),
+			ResourceGroupName:  resourceGroup.Name,
+			VirtualNetworkName: virtualNetwork.Name,
+		})
+		if err != nil {
+			return err
+		}
+
+		subnet3, err := network.NewSubnet(ctx, "subnet-3", &network.SubnetArgs{
```
cnunciato, line 74: Similarly here:

```suggestion
		_, err = network.NewSubnet(ctx, "subnet-3", &network.SubnetArgs{
```

kubernetes-azure-go/main.go
```diff
@@ -0,0 +1,171 @@
...
+	"github.com/pulumi/pulumi/sdk/v3/go/pulumi"
+	"github.com/pulumi/pulumi/sdk/v3/go/pulumi/config"
+)
+
+func main() {
+	pulumi.Run(func(ctx *pulumi.Context) error {
+		// Get some configuration values or set default values
+		cfg := config.New(ctx, "")
+		sshPubKey := cfg.Require("sshPubKey")
+		mgmtGroup := cfg.Require("mgmtGroupId")
```
cnunciato, line 18: Noticed there's a note in the Python version that calls these out. Should we do that here and in the other templates also? (I don't have a opinion other than "consistency is nice". 😄)

```suggestion
		sshPubKey := cfg.Require("sshPubKey")
		mgmtGroup := cfg.Require("mgmtGroupId")
```

kubernetes-azure-yaml/Pulumi.yaml
```diff
@@ -0,0 +1,131 @@
...
+    default: 1.24.3
+  nodeVmSize:
+    type: String
+    default: Standard_DS2_v2
+  mgmtGroupId:
+    type: String
+  sshPubKey:
+    type: String
+resources:
+  # Create a new resource group
```
cnunciato, line 46: Partly personal preference, but I tend to like a little whitespace between sections and things, just for a little breathing room. Just makes it a little easier to see where certain things end and others begin.

cnunciato (approved):
Couple things need fixing up, but otherwise just a handful of non-blocking suggestions. Nice work! ⭐

kubernetes-azure-go/main.go
scottslowe: Yep, this caused an error. Fixed in 8cb24a3.

kubernetes-azure-go/main.go
scottslowe: We definitely need subnet1, but aren't referencing subnet2 (or subnet3) in this template. However, their values might be needed if a user adds more nodepools, so I've added them to the list of exports. Fixed in 8cb24a3.

kubernetes-azure-go/main.go
scottslowe: Fixed via 8cb24a3

kubernetes-azure-go/Pulumi.yaml
scottslowe: Fixed in 965cc5b (reworded to make clear the contents of the key are needed)

kubernetes-azure-go/main.go
scottslowe: Fixed in 965cc5b (comment added to all languages)

kubernetes-azure-yaml/Pulumi.yaml
scottslowe: Fixed in 965cc5b (whitespace added to YAML, already present in other languages)

PR #382, opened 2022-09-27
Always cleanup the temp dirs

### LINKED ISSUE #370: Windows tests failing
The Windows tests have been failing since https://github.com/pulumi/templates/actions/runs/3039081499/jobs/4893593866.

A subset of the AWS tests fail on Windows with this:

```
  Diagnostics:
    aws:s3:Bucket (my-bucket):
      error: unable to validate AWS credentials. Make sure you have:
      
       	 • Set your AWS region, e.g. `pulumi config set aws:region us-west-2`
       	 • Configured your AWS credentials as per https://pulumi.io/install/aws.html
       	 You can also set these via cli using `aws configure`.
```

Oddly, many of the AWS tests (in fact, the ones added in the that first saw these failures) do succeed.

t0yv0:
Likely the CI out of disk space issues occur because we don't clean up the temp folders. https://github.com/pulumi/pulumi/issues/9437 seems to be closed now. Should it be safe to re-enable cleaning the temp folders on CI?

Fixes #370 also

### DIFF
--- a/tests/internal/testutils/environments.go
+++ b/tests/internal/testutils/environments.go
@@ -22,10 +22,6 @@ func NewEnvironment(t *testing.T, cfg TemplateTestConfig) *ptesting.Environment 
 // deleteIfNotFailed deletes the files in the testing environment if the testcase has
 // not failed. (Otherwise they are left to aid debugging.)
 func deleteIfNotFailed(e *ptesting.Environment, cfg TemplateTestConfig) {
-	if cfg.CI {
-		// Skip cleanup on CI, workaround for https://github.com/pulumi/pulumi/issues/9437
-		return
-	}
 	if !e.T.Failed() {
 		e.DeleteEnvironment()
 	}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -95,7 +95,7 @@ jobs:
           aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
           aws-region: ${{ env.AWS_REGION }}
           aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
-          role-duration-seconds: 3600
+          role-duration-seconds: 14400 # 4 hours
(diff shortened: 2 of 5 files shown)


tests/internal/testutils/environments.go
```diff
@@ -22,10 +22,6 @@ func NewEnvironment(t *testing.T, cfg TemplateTestConfig) *ptesting.Environment
 // deleteIfNotFailed deletes the files in the testing environment if the testcase has
 // not failed. (Otherwise they are left to aid debugging.)
 func deleteIfNotFailed(e *ptesting.Environment, cfg TemplateTestConfig) {
-	if cfg.CI {
```
t0yv0, line 25: cfg.CI was set from "CI" env which is set on GitHub actions.

t0yv0:
I've found this was introduced in https://github.com/pulumi/templates/pull/263

t0yv0:
@AaronFriel do we still need the workaround? I can reintroduce it scoped to Windows only. Windows runners seem to have more disk space. However Ubuntu runners are now failing due to out of disk space issues because of this workaround.

Frassle:
windows tests don't seem happy, but the change looks like it _should_ be fine...

t0yv0:
That's interesting. The first time around windows tests simply died (which is indeed suspect that it has something to do with this change). But the second time around (attempt #2) they reported issues:

```
unable to validate AWS credentials.
```

I'm guessing because these now take too long for AWS credentials - the creds expire. This is one hour:

```
          role-duration-seconds: 3600
```

I can extend that.

t0yv0:
Looks like Windows tests die without output on master also https://github.com/pulumi/templates/actions/runs/3137502256/jobs/5095745367

t0yv0:
The latest windows test failure is not related, and is in fact https://github.com/pulumi/templates/issues/386

I recommend we merge this, thoughts?

Frassle (approved):

cnunciato (approved):

PR #383, opened 2022-09-28
Add serverless-azure-* templates

cnunciato:
This PR adds architecture templates for a serverless web app on Azure, modeled on the `serverless-aws-*` templates. It includes templates for TypeScript, Python, Go, C#, and YAML. All languages work except YAML, which should start working once https://github.com/pulumi/pulumi-yaml/pull/342 is released. (Functions are all written in JavaScript for now as well, but I'd like to come back and update these to use the same languages as their containing Pulumi programs at some point soonish.)

I'll probably continue tweaking variable names, comments, and resource names a bit, but for the most part this should be ready to go.


serverless-azure-python/__main__.py
```diff
@@ -0,0 +1,113 @@
...
+site_config = azure.storage.Blob("config.json",
+    account_name=account.name,
+    resource_group_name=resource_group.name,
+    container_name=website.container_name,
+    content_type="application/json",
+    source=app.default_host_name.apply(lambda hostname: pulumi.StringAsset("{ \"api\": \"https://" + hostname  + "/api\" }")))
+
+# Export the URLs of the website and serverless endpoint.
+pulumi.export("originURL", account.primary_endpoints.web)
+pulumi.export("apiURL", app.default_host_name.apply(lambda default_host_name: f"https://{default_host_name}/api"))
```
nimbinatus, line 113: I suspect you could do the `Output.concat()` call here, as well; in this case, I don't think it makes it shorter or easier to read 🤷‍♀️

nimbinatus (commented):
some notes on the Python program

cnunciato:
@nimbinatus I'm closing this in favor of #389 so I can add both serverless-azure and serverless-gcp in the same PR (was just easier to work on them this way), but I'll apply these suggestions over there as well. Thank you for the review!

PR #385, opened 2022-09-28
Upgrade to latest ProgramTest, keep testing against service backend

t0yv0:
Catching up with ProgramTest updates. The new ProgramTest defaults to running with the filestate backend.  This PR keeps continuity so tests and benchmarks keep running on the Service backend.

For benchmarks: it is actually a requirement so we do not observe discontinuity in the timeseries, but I'd like to follow up to run benchmark matrix again with filestate backend to collect that data *also* in addition to service-backend collected data, for comparison.

For tests: debatable whether we should run on Service backend or Filestate. Arguably tests would run a bit faster on Filestate. It's not that we're looking for bugs in the Service backend in these tests.. On the other hand most users would be using the Service backend.


justinvp (approved):
LGTM

PR #387, opened 2022-09-28
feat: validate template configuration against schema and publish JSON

rawkode:


cnunciato:
cc @lukehoban @susanev, here's a look at where we're thinking of going with regard to exposing template metadata to the Service, etc. Right now this is producing an artifact that looks like this:

```json
{
    "templates": [
        {
            "name": "${PROJECT}",
            "description": "${DESCRIPTION}",
            "id": "aiven-go/Pulumi",
            "template": {
                "config": {
                    "aiven:apiToken": {
                        "description": "The token that allows you access to your Aiven account",
                        "secret": true
                    }
                },
                "description": "A minimal Aiven Go Pulumi program"
            },
            "runtime": "go"
        },
        {
            "name": "${PROJECT}",
            "description": "${DESCRIPTION}",
            "id": "alicloud-csharp/Pulumi",
            "template": {
                "config": {
                    "alicloud:region": {
                        "default": "cn-beijing",
                        "description": "The AliCloud region to deploy into"
                    }
                },
                "description": "A minimal AliCloud C# Pulumi program"
            },
            "runtime": "dotnet"
        },
        ...
    ]
}
```

There's more to be done here (adding some fields, tweaking some others), but hopefully this lines up generally with what we've discussed. The idea is we'll build one of these JSON blobs when a change merges to the base branch, and the Service can pick up (and probably cache locally?) the most recent copy at build-time.

* Would merging this PR have

cnunciato:
Hi @rawkode! We ended up going a slightly different route here due to some limitations with the approach we discussed, so we won't be able to merge this after all.

PR #389, opened 2022-09-30
Add serverless-azure-* and serverless-gcp-* templates

cnunciato:
This PR adds architecture templates for a serverless web apps on Azure and Google Cloud Platform, modeled on the `serverless-aws-*` templates. It includes templates for TypeScript, Python, Go, and C#, with serverless functions written in the same languages as their Pulumi programs. 

YAML will follow soon, once https://github.com/pulumi/pulumi-yaml/pull/342 is released (slated for Oct. 5).

Fixes https://github.com/pulumi/devrel-team/issues/174
Fixes https://github.com/pulumi/devrel-team/issues/287


scottslowe (approved):
I've tested all these except the C# versions, and all of them work fine. A few observations:

* For GCP, it appears that users need to have both the Cloud Functions API and the Cloud Build API enabled. It would probably be a good idea to have that information in the landing pages.
* Even though these are Azure and GCP templates, there seems to be an AWS dependency somewhere (I saw it installing AWS providers during the `pulumi new` process). I suspect this is related to the "pulumi-synced-folder" component.
* Also possibly related to "pulumi-synced-folder" is that it seems to install an older version of Azure Native (1.71.0), even when a newer version is already installed.

Aside from these observations, the templates work as advertised and the UX is simple and straightforward.

cnunciato:
Excellent, thanks for the thorough review @scottslowe. Good callout on the GCP requirements, too; we'll make sure to get those documented as well (cc @gthuang).

> there seems to be an AWS dependency somewhere

That's interesting and good to know. My understanding's so far been that those unused providers wouldn't be downloaded, but maybe there's something about how they're being referenced somewhere that's triggering that. It might also just make sense to break this component up into separate components; I did this purely for expediency for the static-website templates, but in practice it might just heap more weight onto the user.

> it seems to install an older version of Azure Native

Indeed, and I believe this is just "how it works" in that that's the version the provider was compiled against. It'd probably be good to have these components be rebuilt when new versions of their dependencies are released, though. Probably worth filing an issue on this one.

cnunciato:
YAML's been added to this PR, so we have all five languages now.

PR #391, opened 2022-10-02
AWS VM templates

aaronkao:
PR for https://github.com/pulumi/devrel-team/issues/448


scottslowe:
I assume this is all using the default VPC? Is it worth adding the ability for the user to use a different VPC?

aaronkao:
> I assume this is all using the default VPC? Is it worth adding the ability for the user to use a different VPC?

yeah that probably makes sense. what is the method to ask the user to pass in which VPC? just the VPC ID or ARN?

scottslowe:
> what is the method to ask the user to pass in which VPC? just the VPC ID or ARN?
I'd say you have a couple of options. One, you could use a "get VPC"-type function to look up a VPC, including the default VPC or some other VPC. Alternately, you could have the user supply the VPC ID via a configuration value.

joeduffy:
We do get flak from time to time for assuming people have default VPCs. Most major enterprises have a rule to disallow default VPCs, which means many examples/etc simply don’t work for folks. I think even worse than us assuming this is the poor failure mode which leaves people left with unfriendly errors and an unclear path forward. Even if it’s just a comment that says “this assumes a default VPC and if you don’t have one, do XYZ”, it would probably go a long way. (Btw would also be good to develop a general philosophy and guidance around this.)

cnunciato:
I might just create a new VPC here, like we do in this example: https://www.pulumi.com/docs/aws/ec2/ Any reason not to do that?

scottslowe (dismissed):
Given that we are close to announcing AWSX GA (as I understand it), it might be better to use AWSX here to create the VPC and associated infrastructure instead of creating all of it separately.

aaronkao:
Bug https://github.com/pulumi/pulumi-awsx/issues/931 is preventing me from using AWSX, going to go forward with just standard AWS library

vm-aws-yaml/Pulumi.yaml
```diff
@@ -0,0 +1,124 @@
...
+            
+            echo \"Hello, World from Pulumi!\" > index.html
+            
+            nohup python -m SimpleHTTPServer 80 &"
+            
+
+outputs:
+    # Export the public IP and DNS
+    publicIP: ${server.publicIp}
+    publicDNS: ${server.publicDns}
```
cnunciato, line 124: I've generally been calling these exports just `ip` and `hostname`. Ok with those, just for consistency?

```suggestion
    ip: ${server.publicIp}
    hostname: ${server.publicDns}
    url: http://${${server.publicDns}}
```

Adding `url` is also nice because you can immediately click it to open a browser, and on the template pages, we can do things like we do here, with `open $(pulumi stack output url)`, for a bit more engagement: https://www.pulumi.com/templates/container-service/aws/

vm-aws-yaml/Pulumi.yaml
```diff
@@ -0,0 +1,124 @@
...
+            return: id    
+    # User data to start a HTTP server in the EC2 instance
+    userData: "#!/bin/bash
+            
+            echo \"Hello, World from Pulumi!\" > index.html
+            
+            nohup python -m SimpleHTTPServer 80 &"
+            
+
+outputs:
```
cnunciato, line 121:
```suggestion
    userData: "#!/bin/bash
        echo \"Hello, World from Pulumi!\" > index.html
        nohup python -m SimpleHTTPServer 80 &"
            
outputs:
```

vm-aws-yaml/Pulumi.yaml
```diff
@@ -0,0 +1,124 @@
+name: vm-aws-yaml
+description: A YAML program to create a virtual machine on AWS
```
cnunciato, line 2: To allow these names to be set at `pulumi new`-time, we use these expressions:

```suggestion
name: ${PROJECT}
description: ${DESCRIPTION}
```

vm-aws-yaml/Pulumi.yaml
```diff
@@ -0,0 +1,124 @@
...
+runtime: yaml
+
+template: 
+    description: A YAML program to create a virtual machine on AWS
```
cnunciato, line 6: I think we wanted to remove the language names for these, right? 

```suggestion
    description: A virtual machine on Amazon EC2
```

vm-aws-yaml/Pulumi.yaml
```diff
@@ -0,0 +1,124 @@
...
+    config:
+        aws:region:
+            description: The AWS region to deploy into
+            default: us-west-2
+        instanceType:
+            description: Instance type of EC2
+            default: t2.micro
+        vpcNetworkCidr:
+            description: Network CIDR to use for new VPC
```
cnunciato, line 15:
```suggestion
            description: The network CIDR to use for the VPC
```

vm-aws-yaml/Pulumi.yaml
```diff
@@ -0,0 +1,124 @@
...
+            default: 10.0.0.0/16
+
+configuration:
```
cnunciato, line 18: It's a remnant of the YAML launch, but basically everything from here down needs to be repeated in a new file called `Pulumi.yaml.append` (and needs to match what's in this file of course). You'll see this in other YAML templates as well if you want some examples. We should be able to remove these soon, though: https://github.com/pulumi/templates/issues/325

vm-aws-yaml/Pulumi.yaml
```diff
@@ -0,0 +1,124 @@
...
+            cidrBlock: 10.0.0.0/16
+            enableDnsHostnames: true
+            enableDnsSupport: true 
+    # Create an internet gateway
+    gateway:
+        type: aws:ec2:InternetGateway
+        properties:
+            vpcId: ${vpc.id}
+            tags:
+                Name: vm-aws-yaml-ig
```
cnunciato, line 40: Do these need to be tagged? Would users expect them to be?

vm-aws-yaml/Pulumi.yaml
```diff
@@ -0,0 +1,124 @@
...
+configuration:
+    instanceType:
+        type: String
+        default: t3.micro
+    vpcNetworkCidr:
+        type: String
+        default: 10.0.0.0/16
+
+resources:  
+    # Create VPC
```
cnunciato, line 27:
```suggestion
    # Create VPC.
```

vm-aws-yaml/Pulumi.yaml
```diff
@@ -0,0 +1,124 @@
...
+    vpc:
+        type: aws:ec2:Vpc
+        properties:
+            cidrBlock: 10.0.0.0/16
+            enableDnsHostnames: true
+            enableDnsSupport: true 
+    # Create an internet gateway
```
cnunciato, line 34:
```suggestion
    # Create an internet gateway.
```

vm-aws-yaml/Pulumi.yaml
```diff
@@ -0,0 +1,124 @@
...
+    # Create a subnet that automatically assigns new instances a public IP address.
+    subnet:
+        type: aws:ec2:Subnet
+        properties:
+            vpcId: ${vpc.id}
+            cidrBlock: 10.0.1.0/24
+            mapPublicIpOnLaunch: true
+            tags:
+                Name: vm-aws-yaml-subnet
+    # Create a route table
```
cnunciato, line 50:
```suggestion
    # Create a route table.
```

vm-aws-yaml/Pulumi.yaml
```diff
@@ -0,0 +1,124 @@
...
+                  gatewayId: ${gateway.id}
+            tags:
+                Name: vm-aws-yaml-routetable
+    # Associate the route table with the public subnet.
+    routeTableAssociation:
+        type: aws:ec2:RouteTableAssociation
+        properties:
+            subnetId: ${subnet.id}
+            routeTableId: ${routeTable.id}
+    # Create EC2 Security Group
```
cnunciato, line 66: I think it's useful to be a little more verbose on this one, just for clarity:

```suggestion
    # Create a security group allowing inbound access over port 80 and outbound
    # access to anywhere.
```

vm-aws-yaml/Pulumi.yaml
```diff
@@ -0,0 +1,124 @@
...
+            subnetId: ${subnet.id}
+            vpcSecurityGroupIds: 
+                - ${secGroup}
+            userData: ${userData}
+            ami: ${ami}
+            tags:
+                Name: vm-aws-yaml-webserver
+
+variables:
+    # Create EC2 AMI
```
cnunciato, line 100:
```suggestion
    # Look up the latest Amazon Linux 2 AMI.
```

vm-aws-yaml/Pulumi.yaml
```diff
@@ -0,0 +1,124 @@
...
+                    - 0.0.0.0/0
+            egress:
+                - fromPort: 0
+                  toPort: 0
+                  protocol: -1
+                  cidrBlocks:
+                    - 0.0.0.0/0
+            tags:
+                Name: vm-aws-yaml-secgrp
+    # Create EC2 server
```
cnunciato, line 86:
```suggestion
    # Create and launch an EC2 instance into the public subnet.
```

vm-aws-yaml/Pulumi.yaml
```diff
@@ -0,0 +1,124 @@
...
+                mostRecent: true
+            return: id    
+    # User data to start a HTTP server in the EC2 instance
+    userData: "#!/bin/bash
+            
+            echo \"Hello, World from Pulumi!\" > index.html
+            
+            nohup python -m SimpleHTTPServer 80 &"
+            
+
```
cnunciato, line 120:
```suggestion
# Export the instance's publicly accessible IP address and hostname.
```

cnunciato (commented):
Bunch of little updates here, but this is looking good!

My first question after I do this is going to be "How do I SSH into this?" Which don't think I can do, because I didn't specify an SSH key file. Maybe that's something we can add as a next step, though. What do you think?

vm-aws-yaml/Pulumi.yaml
```diff
@@ -0,0 +1,126 @@
...
+outputs:
+    # Export the public IP and DNS
```
aaronkao, line 123:
```suggestion
```

vm-aws-python/__main__.py
```diff
@@ -0,0 +1,87 @@
...
+vpc_network_cidr = config.get("vpcNetworkCidr")
+if vpc_network_cidr is None:
+    vpc_network_cidr = "10.0.0.0/16"
+
+# Look up the latest Amazon Linux 2 AMI.
+ami = aws.ec2.get_ami(filters=[aws.ec2.GetAmiFilterArgs(
+        name="name",
+        values=["amzn-ami-hvm-*"],
+    )],
+    owners=["137112412989"],
```
joeduffy, line 18: I have found that using `"amazon"` is much more intuitive versus `"137112412989"`. (They resolve to the same.)

vm-aws-python/__main__.py
```diff
@@ -0,0 +1,87 @@
...
+if instance_type is None:
+    instance_type = "t3.micro"
+vpc_network_cidr = config.get("vpcNetworkCidr")
+if vpc_network_cidr is None:
+    vpc_network_cidr = "10.0.0.0/16"
+
+# Look up the latest Amazon Linux 2 AMI.
+ami = aws.ec2.get_ami(filters=[aws.ec2.GetAmiFilterArgs(
+        name="name",
+        values=["amzn-ami-hvm-*"],
```
joeduffy, line 16: I believe (but may be mistaken) that `"amzn-ami-hvm-*"` resolves to Amazon Linux 1, not 2. I believe `"amzn2-ami-hvm-*"` resolves to Amazon Linux 2. Not 100% sure here but I was bitten by this recently.

vm-aws-python/__main__.py
aaronkao: Done

vm-aws-python/__main__.py
aaronkao: Done

vm-aws-csharp/{$PROJECT}.csproj
```diff
@@ -0,0 +1,14 @@
+<Project Sdk="Microsoft.NET.Sdk">
```
cnunciato, line 1: This file should be named `${PROJECT}.csproj` (with the `$` on the outside).

vm-aws-go/go.mod
```diff
@@ -0,0 +1,58 @@
+module vm-aws-yaml
```
cnunciato, line 1:
```suggestion
module ${PROJECT}
```

vm-aws-go/Pulumi.yaml
```diff
@@ -0,0 +1,15 @@
+name: ${PROJECT}
+runtime: go
+description: ${DESCRIPTION}
```
cnunciato, line 3:
```suggestion
name: ${PROJECT}
description: ${DESCRIPTION}
runtime: go
```

vm-aws-python/Pulumi.yaml
```diff
@@ -0,0 +1,15 @@
+name: ${PROJECT}
+runtime: python
+description: ${DESCRIPTION}
```
cnunciato, line 3:
```suggestion
name: ${PROJECT}
description: ${DESCRIPTION}
runtime: python
```

vm-aws-typescript/Pulumi.yaml
```diff
@@ -0,0 +1,15 @@
+name: ${PROJECT}
+runtime: nodejs
+description: ${DESCRIPTION}
```
cnunciato, line 3:
```suggestion
name: ${PROJECT}
description: ${DESCRIPTION}
runtime: nodejs
```

vm-aws-typescript/index.ts
```diff
@@ -0,0 +1,91 @@
...
+});
+
+// Associate the route table with the public subnet.
+const routeTableAssociation = new aws.ec2.RouteTableAssociation("routeTableAssociation", {
+    subnetId: subnet.id,
+    routeTableId: routeTable.id,
+});
+
+/* Create a security group allowing inbound access over port 80 and outbound
+access to anywhere. */
```
cnunciato, line 58:
```suggestion
// Create a security group allowing inbound access over port 80 and outbound
// access to anywhere.
```

vm-aws-typescript/package.json
```diff
@@ -0,0 +1,11 @@
+{
+		"name": "vm-aws-yaml",
+		"devDependencies": {
+			"@types/node": "^14"
+		},
+		"dependencies": {
+			"typescript": "^4.0.0",
+			"@pulumi/pulumi": "^3.0.0",
+			"@pulumi/aws": "5.18.0"
+		}
```
cnunciato, line 10:
```suggestion
    "name": "${PROJECT}",
    "devDependencies": {
        "@types/node": "^14"
    },
    "dependencies": {
        "typescript": "^4.0.0",
        "@pulumi/pulumi": "^3.0.0",
        "@pulumi/aws": "5.18.0"
    }
```

vm-aws-typescript/tsconfig.json
```diff
@@ -0,0 +1,18 @@
...
+			"sourceMap": true,
+			"experimentalDecorators": true,
+			"pretty": true,
+			"noFallthroughCasesInSwitch": true,
+			"noImplicitReturns": true,
+			"forceConsistentCasingInFileNames": true
+		},
+		"files": [
+			"index.ts"
+		]
```
cnunciato, line 17:
```suggestion
    "compilerOptions": {
        "strict": true,
        "outDir": "bin",
        "target": "es2016",
        "module": "commonjs",
        "moduleResolution": "node",
        "sourceMap": true,
        "experimentalDecorators": true,
        "pretty": true,
        "noFallthroughCasesInSwitch": true,
        "noImplicitReturns": true,
        "forceConsistentCasingInFileNames": true
    },
    "files": [
        "index.ts"
    ]
```

vm-aws-csharp/Program.cs
```diff
@@ -0,0 +1,141 @@
...
+
+    // Associate the route table with the public subnet.
+    var routeTableAssociation = new Aws.Ec2.RouteTableAssociation("routeTableAssociation", new()
+    {
+        SubnetId = subnet.Id,
+        RouteTableId = routeTable.Id,
+    });
+
+    /* Create a security group allowing inbound access over port 80 and outbound
+    access to anywhere. */
```
cnunciato, line 83:
```suggestion
    // Create a security group allowing inbound access over port 80 and outbound
    // access to anywhere.
```

vm-aws-csharp/Pulumi.yaml
```diff
@@ -0,0 +1,15 @@
+name: ${PROJECT}
+runtime: dotnet
+description: ${DESCRIPTION}
```
cnunciato, line 3:
```suggestion
name: ${PROJECT}
description: ${DESCRIPTION}
runtime: dotnet
```

cnunciato (commented):
All work as expected. 🎉  Bunch of things we should fix up before merging, but otherwise 👍

cnunciato (approved):

PR #394, opened 2022-10-04
Add Kubernetes on GCP templates

scottslowe:
This PR adds templates to stand up Kubernetes on GCP (GKE) in Go, Python, TypeScript, and YAML. C# and Java will be added later via a separate PR.


timbohiatt:
Hey @scottslowe and team;

Just had a look at this one for you all. I would love to know about the GCP Project you use to test the deployment of these templates generally? One of the big differentiators in GCP vs some other providers is the ability to at certain levels of the Organization Hierarchy enable or disable [Service API's](https://cloud.google.com/apis/docs/overview). This can be done via policies from the org -> folder -> project level. But actual enablement of each service API is done at the Project Level. If an API is not enabled I expect Pulumi will fail on ```pulumi up```.

If you are using a GCP project for testing where the required API's for a Pulumi deployment have already been enabled. I believe this would mean a tester not using a NEW GCP project may never see this error. But someone who has a brand new GCP project and deploys the template may get tripped up. 

Possibly we should look to ensure all Pulumi Templates Enable the Required API's for all services used as part of the template to reduce friction. Alternatively this information should be a prerequisite in the Pulumi/Templates/Hugo-Docs?

**Service API Docs:**
_GCP Docs:    https://cloud.google.com/apis/docs/overview
Pulumi Docs: https://www.pulumi.com/registry/packages/gcp/api-docs/projects/service/_

Following this comment I will test each of these templates from a scratch GCP project and double check my assumptions here.

timbohiatt:
Attempted to deploy this template directly from your PullRequest to a fresh GCP project and it confirmed my initial thinking. 
In any new GCP project the API's are not enabled. In this case initially the api required is ```compute.googleapis.com```

I ran the pulumi new command as follows and walked through the teamplate configuration. When asking for the GCP project I provided a new GCP project:
```pulumi new https://github.com/pulumi/templates/tree/slowe/kubernetes-gcp-templates/kubernetes-gcp-python```

I also attempted to deploy it in an existing project where I had a lot of API's that were already enabled and it deployed with no problems at all. 

Included here is the error you get when the API's are not enabled in a project. 
image

Before we made contact I was already having a go at making a [GCP template for containers](https://github.com/timbohiatt/pulumi-templates/tree/google-classic/ContainerService/container-gcp-typescript) I held off raising a PR on this as you mentioned yourself and @cnunciato were already working on it. However in that PR there is an [example of enabling the API's lines 11-28 as an example if it is helpful](https://github.com/timbohiatt/pulumi-templates/blob/google-classic/ContainerService/container-gcp-typescript/index.ts#L11)

**Comment by scottslowe** [comment]
@timbohiatt Our initial thought was that we would document (in the template's associated landing page on the Pulumi web site) the requirement to have certain APIs enabled. However, I'm certainly open to programmatically doing it instead. I'll review the link and see if we want to add those to the templates or document the requirements for enabling the APIs before running `pulumi up`.

**Comment by timbohiatt** [comment]
@scottslowe both are great options really. 

From our side when we talk about content with Terraform specifically with customers we try to encourage the API activation in the IaC from Day 0. It increases the ability to move a PoC to a new project and use CI/CD processes going forward as well as repeatability. One of things we like about Pulumi's idea behind published templates is; Many of our customers are trying to do the same with "building block" blueprints. Where their users can just deploy this blueprint + another blue print and string them together. API activation in each template helps people who have IaC deployment access and API access achieve this even when they cannot access the console or need to provide accountability/auditing as part of their IaC processes. 

As funny as it may seem we often see heavily regulated customers who can run API's but can't access cloud console to enable api's with the UI. This is strange but true;

**Comment by scottslowe** [comment]
Thanks for that context! In all likelihood, we'll go the route of activating the APIs in the code, but I just want to make sure I have consensus among everyone involved.

**Comment by timbohiatt** [comment]
That makes perfect sense! 
Team thinking is the only way to scale correctly :)

File: kubernetes-gcp-go/main.go [hunk 1]
```diff
@@ -0,0 +1,152 @@
...
+				WorkloadPool: pulumi.String(fmt.Sprintf("%v.svc.id.goog", gcpProject)),
+			},
+		})
+		if err != nil {
+			return err
+		}
+
+		// Create a GCP Service Account for the node pool
+		gkeNodepoolSa, err := serviceaccount.NewAccount(ctx, "gke-nodepool-sa", &serviceaccount.AccountArgs{
+			AccountId:   pulumi.String("nodepool-1-sa"),
```
cnunciato, line 96: I'm not sure what the best way around this is, but it looks like this string has to be unique within the scope of a GCP project; I only noticed it because I tried to stand up a TypeScript instance having already created one in YAML. 

It'll probably cause problems for us in test, but that aside, once two people using the same GCP project try to use this template, they'll run into the same issue. I wonder if using the Random provider would help us around it?

kubernetes-gcp-go/main.go
```diff
@@ -0,0 +1,152 @@
...
+			DisplayName: pulumi.String("Nodepool 1 Service Account"),
```
cnunciato, line 97: If we do choose to change the way we name the account ID, we might want to make this something more generic (or just leave it out, unless that makes things too weird in the GCP console).

kubernetes-gcp-go/main.go
```diff
@@ -0,0 +1,152 @@
...
+kind: Config
+preferences: {}
+users:
+- name: %[1]s
+  user:
+    exec:
+      apiVersion: client.authentication.k8s.io/v1beta1
+      command: gke-gcloud-auth-plugin
+      installHint: Install gke-gcloud-auth-plugin for use with kubectl by following
+        https://cloud.google.com/blog/products/containers-kubernetes/kubectl-auth-changes-in-gke
```
cnunciato, line 140: Interesting -- how does this work?

kubernetes-gcp-python/__main__.py
```diff
@@ -0,0 +1,117 @@
+import pulumi
+import pulumi_gcp as gcp
+
+# Get some provider-namespaced configuration values
+provider_cfg = pulumi.Config("gcp")
+gcp_project = provider_cfg.require("project")
+gcp_region = provider_cfg.get("region")
+if gcp_region is None:
+    gcp_region = "us-central1"
```
cnunciato, line 9: Should we use Laura's default-value thing here, and with the others?

```suggestion
gcp_region = provider_cfg.get("region", "us-central1")
```

kubernetes-gcp-typescript/package.json
```diff
@@ -0,0 +1,11 @@
...
+		"name": "tmp",
+		"devDependencies": {
+			"@types/node": "^14"
+		},
+		"dependencies": {
+			"typescript": "^4.0.0",
+			"@pulumi/pulumi": "^3.0.0",
+			"@pulumi/gcp": "6.39.0"
+		}
+}
```
cnunciato, line 11:
```suggestion
{
    "name": "tmp",
    "devDependencies": {
        "@types/node": "^14"
    },
    "dependencies": {
        "typescript": "^4.0.0",
        "@pulumi/pulumi": "^3.0.0",
        "@pulumi/gcp": "6.39.0"
    }
}
```

kubernetes-gcp-typescript/tsconfig.json
```diff
@@ -0,0 +1,18 @@
...
+			"experimentalDecorators": true,
+			"pretty": true,
+			"noFallthroughCasesInSwitch": true,
+			"noImplicitReturns": true,
+			"forceConsistentCasingInFileNames": true
+		},
+		"files": [
+			"index.ts"
+		]
+}
```
cnunciato, line 18:
```suggestion
{
    "compilerOptions": {
        "strict": true,
        "outDir": "bin",
        "target": "es2016",
        "module": "commonjs",
        "moduleResolution": "node",
        "sourceMap": true,
        "experimentalDecorators": true,
        "pretty": true,
        "noFallthroughCasesInSwitch": true,
        "noImplicitReturns": true,
        "forceConsistentCasingInFileNames": true
    },
    "files": [
        "index.ts"
    ]
}
```

kubernetes-gcp-typescript/package.json
cnunciato: (There's an [open issue](https://github.com/pulumi/pulumi/issues/10776) on the codegen for this.)

cnunciato (commented):
A few things to consider that span all languages. Main thing, though, is that we'll need to figure out what to do about those account IDs. 🤔

kubernetes-gcp-go/main.go
timbohiatt: GCP service accounts do have to be unique that's correct! 

Would it be possible to use the GetAccount method from the Pulumi provider to determine if an account first exists. If it does use the existing account; should the "GetAccount" method return null/nil/undefined/ect then run the create statement? 

Typically gcp recommended best practice would be one service account per process scoped exactly with only the permissions required. Therefore if you take this approach you could add the unique cluster name into to the SA AcountId? Therefore if you made two clusters with different names you would get two cluster SAs with different ids? 

In the real word we try to steal away from uuid/random postfixes and normally recommend customers use a controlled increment.

scottslowe: Embedding the cluster name (which _should_ have a unique name due to Pulumi's auto-naming mechanisms) would take care of the issue with the service account, I'd guess.

kubernetes-gcp-go/main.go
timbohiatt: I have never actually seen it happen but i believe it's the error message that gets shown to a user if the don't have ```gke-gcloud-auth-plugin``` on the system.  Not actually sure how it will work in a Pulumi run though.

kubernetes-gcp-python/__main__.py
timbohiatt: The use of .get is significantly neater.

kubernetes-gcp-go/main.go
```diff
@@ -0,0 +1,152 @@
...
+				},
+				ServiceAccount: gkeNodepoolSa.Email,
+			},
+		})
+		if err != nil {
+			return err
+		}
+
+		// Build Kubeconfig for accessing the cluster
+		clusterKubeconfig := pulumi.Sprintf(`apiVersion: v1
```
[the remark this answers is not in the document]
timbohiatt: helps with both readability. If a Pulumi user later takes this template and uses as the foundation for creating multiple GKE clusters or a "Fleet of GKE" clusters. Having this code in a function would make scalability simpler.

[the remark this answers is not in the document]
scottslowe: @cnunciato Do you happen to have an example of how I could/would pull this into a function?

kubernetes-gcp-go/Pulumi.yaml
```diff
@@ -0,0 +1,14 @@
...
+runtime: go
+template:
+  description: A Go program to create a Kubernetes cluster on GCP
+  config:
+    gcp:project:
+      description: The GCP project in which to create resources
+    gcp:region:
+      default: us-central1
+      description: The GCP region in which resources should be created
+    nodePoolCount:
```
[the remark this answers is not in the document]
scottslowe: I've switched it to `nodesPerZone` as of commit 440fb15

kubernetes-gcp-go/Pulumi.yaml
```diff
@@ -0,0 +1,14 @@
...
+description: ${DESCRIPTION}
+runtime: go
+template:
+  description: A Go program to create a Kubernetes cluster on GCP
+  config:
+    gcp:project:
+      description: The GCP project in which to create resources
+    gcp:region:
+      default: us-central1
+      description: The GCP region in which resources should be created
```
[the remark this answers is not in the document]
scottslowe: Updated in 440fb15

kubernetes-gcp-go/main.go
scottslowe: In 440fb15, I added the uniquely-named cluster name to the name of the node pool SA, which guarantees unique-ness.

kubernetes-gcp-python/__main__.py
scottslowe: Good catch! Fixed in 440fb15.

kubernetes-gcp-typescript/package.json
scottslowe: Fixed in 440fb15

kubernetes-gcp-typescript/tsconfig.json
scottslowe: Fixed in 440fb15

kubernetes-gcp-go/main.go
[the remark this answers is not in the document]
cnunciato: Sorry for the delayed response -- not at the ready, but I'll see if i can get something for you. Not need to block on it, though; it's definitely the kind of thing that can be added as a follow-up.

kubernetes-gcp-python/__main__.py
```diff
@@ -0,0 +1,113 @@
...
+
+# Create a new network
+gke_network = gcp.compute.Network("gke-network",
+    auto_create_subnetworks=False,
+    description="A virtual network for your GKE cluster(s)"
+)
+
+# Create a subnet in the new network
+gke_subnet = gcp.compute.Subnetwork("gke-subnet",
+    ip_cidr_range="10.128.0.0/12",
```
nimbinatus, line 20: Nit for PEP8

```suggestion
gke_subnet = gcp.compute.Subnetwork(
    "gke-subnet",
    ip_cidr_range="10.128.0.0/12",
```

kubernetes-gcp-python/__main__.py
```diff
@@ -0,0 +1,113 @@
...
+    network=gke_network.id,
+    private_ip_google_access=True
+)
+
+# Create a cluster in the new network and subnet
+gke_cluster = gcp.container.Cluster("gke-cluster",
+    addons_config=gcp.container.ClusterAddonsConfigArgs(
```
nimbinatus, line 27: Nit for PEP8
```suggestion
gke_cluster = gcp.container.Cluster(
    "gke-cluster",
    addons_config=gcp.container.ClusterAddonsConfigArgs(
```

kubernetes-gcp-python/__main__.py
```diff
@@ -0,0 +1,113 @@
...
+    ),
+    subnetwork=gke_subnet.name,
+    workload_identity_config=gcp.container.ClusterWorkloadIdentityConfigArgs(
+        workload_pool=f"{gcp_project}.svc.id.goog"
+    )
+)
+
+# Create a GCP service account for the nodepool
+gke_nodepool_sa = gcp.serviceaccount.Account("gke-nodepool-sa",
+    account_id=pulumi.Output.concat(gke_cluster.name, "-np-1-sa"),
```
nimbinatus, line 68: Nit for PEP8
```suggestion
gke_nodepool_sa = gcp.serviceaccount.Account(
    "gke-nodepool-sa",
    account_id=pulumi.Output.concat(gke_cluster.name, "-np-1-sa"),
```

kubernetes-gcp-python/__main__.py
```diff
@@ -0,0 +1,113 @@
...
+    display_name="Nodepool 1 Service Account"
+)
+
+# Create a nodepool for the cluster
+gke_nodepool = gcp.container.NodePool("gke-nodepool",
+    cluster=gke_cluster.id,
```
nimbinatus, line 74: Nit for PEP8

```suggestion
gke_nodepool = gcp.container.NodePool(
    "gke-nodepool",
    cluster=gke_cluster.id,
```

kubernetes-gcp-python/__main__.py
```diff
@@ -0,0 +1,113 @@
...
+    node_count=nodes_per_zone,
+    node_config=gcp.container.NodePoolNodeConfigArgs(
+        oauth_scopes=["https://www.googleapis.com/auth/cloud-platform"],
+        service_account=gke_nodepool_sa.email
+    )
+)
+
+# Build a Kubeconfig to access the cluster
+cluster_kubeconfig = pulumi.Output.all(gke_cluster.master_auth.cluster_ca_certificate, gke_cluster.endpoint, gke_cluster.name).apply(lambda l: f"""apiVersion: v1
```
nimbinatus, line 83: PEP8/linters on a typical IDE probably will throw errors or warnings on this line, depending on people's settings. The only way I know to fix it is to move all the Outputs to their own line :/

nimbinatus (commented):
A couple PEP8 nits on the Python ones that are up to you whether you approve. The Python stuff looks great, though! And thank you for putting the arguments in alphabetical order!

kubernetes-gcp-python/__main__.py
scottslowe: Fixed in fa13861

kubernetes-gcp-python/__main__.py
scottslowe: Fixed in fa13861

kubernetes-gcp-python/__main__.py
scottslowe: Fixed in fa13861

kubernetes-gcp-python/__main__.py
scottslowe: Fixed in fa13861

kubernetes-gcp-python/__main__.py
scottslowe: Fixed in fa13861 (moved each Output to its own line and ran it through an online PEP8 linter, which didn't report any issues)

kubernetes-gcp-go/main.go
cnunciato: I see this myself, yeah. I'm using `brew` to manage my `gcloud` binary, so it looks like `gke-gcloud-auth-plugin` gets installed in the same directory as `gcloud`, but because only `gcloud` is symlinked, it's is not on my path. When I tweak my path manually to include `gke-gcloud-auth-plugin`, though, the message goes away. 

Unfortunately the linked article doesn't address the Homebrew-installed use case, so I had to fumble around looking for `gke-gcloud-auth-plugin` on my system to figure it out. I'm assuming that probably means most people don't install `gcloud` with Brew.

kubernetes-gcp-go/main.go
```diff
@@ -0,0 +1,152 @@
...
+        https://cloud.google.com/blog/products/containers-kubernetes/kubectl-auth-changes-in-gke
+      provideClusterInfo: true
+        `, gkeCluster.Name, gkeCluster.Endpoint, gkeCluster.MasterAuth.ClusterCaCertificate().Elem())
+
+		// Export some values for use elsewhere
+		ctx.Export("networkName", gkeNetwork.Name)
+		ctx.Export("networkId", gkeNetwork.ID())
+		ctx.Export("clusterName", gkeCluster.Name)
+		ctx.Export("clusterId", gkeCluster.ID())
+		ctx.Export("kubeconfig", clusterKubeconfig)
```
cnunciato, line 149: We're probably planning on mentioning this in the landing page, but figured I'd call it out just in case. My first attempt to output `kubeconfig` resulted in a file containing only `[secret]`. I had to pass `--show-secrets`:

```
pulumi stack output --show-secrets kubeconfig > kubeconfig
```

.. to get it to work.

nimbinatus (approved):
:shipit: for the Python one!

kubernetes-gcp-go/main.go
timbohiatt: Out of curiosity; Does this only occur with A) the Go implementation and B) does the same level of configuration occur with Azure and AWS implementations?

scottslowe: @timbohiatt AWS and Azure behave differently; the Kubeconfig doesn't have to be built manually, like we do here for GKE, and the export/output isn't treated like a secret. I'll dig into this code to see if I can figure out _why_ the Kubeconfig is being treated like a secret.

cnunciato (approved):
Everything works and LGTM. ⭐ 

The only thing I noticed worth mentioning while running through these is that the `kubeconfig` output is treated differently in YAML vs. in other languages. In YAML, that value is exported as a plain string, but in the other languages, it's marked as a secret -- which means that to get at the value, you need to `export` with `--show-secrets`. (So the docs would presumably need to reflect this somehow.) We'll probably want to figure this out before merging, but otherwise, what's here looks great to me.

kubernetes-gcp-go/main.go
timbohiatt: Yeah this is actually what I thought you might say; I would be interested in your finding around this. I wonder if it's to do with the use how the GCP versions maybe uses the raw k8's connector and therefore requires the actual kubeconfig object to be built. We find this is a lot with GCP that typically we don't create "wrappers" around K8's standard content. I would be keen to see what you find out.

timbohiatt (approved):

cnunciato:
Also one thing @scottslowe -- you'll need to add these a new item (`kubernetes-gcp`) under `metadata/groups` as well. You can model it on `kubernetes-aws` and list only the templates we've written (so leave `csharp`, etc.)

https://github.com/pulumi/templates/tree/master/metadata/groups

You can do that either in this PR or as a follow-up, up to you. (I still haven't written this into these docs, but will do ASAP.)

kubernetes-gcp-go/main.go
scottslowe: I haven't gotten to the root cause of why the Go, Python, and TypeScript implementations treat the Kubeconfig as a secret, but for consistency's sake I've modified the YAML implementation so that it behaves similarly. This will at least give us consistency across languages within this set of templates. I'll push that change shortly.

scottslowe:
@cnunciato I've updated the YAML version to treat the Kubeconfig output as a secret for consistency with the other languages. I'll update the template metadata in a separate PR. Thanks!

PR #395, opened 2022-10-05
Add support for template metadata

cnunciato:
This change adds support for generating a JSON list of templates and template architectures and storing the result as a GitHub release artifact. Also contains tests to ensure the data remains correct.

To generate the list and run the tests, run:

```bash
make metadata
```

To fetch the a built artifact from GitHub (once this change has been merged to master):

```bash
curl -L https://github.com/pulumi/templates/releases/download/latest/metadata.json
```

Fixes https://github.com/pulumi/devrel-team/issues/438.


susanev (commented):
we do not have to fix this now especially because we arent exposing descriptions in the service ui quite yet, but when we do we may want to revisit these descriptions. they all seem to be almost exactly the same except for the part where we repeat the template name. so im not sure how much value there is to them for the service ui and also not sure if they will help people decide on a template. i wonder if we can provide more deets in the description about the template, maybe the resources it will create, or config it requires, or other details that are unique to each template.

susanev:
@sean1588 can you link to the issue you are working from for @cnunciato? i believe Christian built this out assuming we were building out the larger experience and not just expositing the dropdown with the templates. might be good for him to see that to help inform things here.

sean1588:
here is the issue: https://github.com/pulumi/pulumi-service/issues/9508

cnunciato:
@sean1588 @susanev All right, I'm going to remove these descriptions for now, then; we can always add them back in later. I'll leave the list of templates, though, as I suspect you'll be needing those sooner than later.

cnunciato:
Hold up -- looks like I have a few Git things to reconcile first..

scripts/build-metadata.js
```diff
@@ -0,0 +1,53 @@
...
+    });
+
+// Assemble the list.
+const result = {
+    architectures: architectureTemplates,
+    templates: allTemplates,
+};
+
+// Write the list to an output file.
+fs.writeFileSync("metadata.json", JSON.stringify(result, null, 4), "utf-8");
```
sean1588, line 53: Any chance we could have some sort of check to validate the generated metadata hasn't structurally drifted. I see you added a test stage to the workflow, but not sure if it validates this. If not, or if it would be too much work, would it be possible to just have some sort of check where people need to add a label to verify they didn't touch this file when opening a PR against this repo, similar to what we do in the service repo with the changelog file. Not sure if you have any other ideas around this.

metadata/groups/static-website-azure.yaml
```diff
@@ -0,0 +1,12 @@
+name: Azure Static Website
+kind: architecture
+parent: static-website
+slug: azure
+clouds:
+  - azure
```
sean1588, line 6: I'm guessing some time down the road it would be possible that there will be groupings that are associated with multiple clouds?

cnunciato: I was thinking that might be possible, yeah (though certainly not common), so just figured I'd leave room for it.

scripts/build-metadata.js
cnunciato: The tests (which run on PR and push events) do verify structure, yeah, and we won't update the release unless those tests pass.

>would it be possible to just have some sort of check where people need to add a label to verify they didn't touch this file

I'm not sure I totally understand -- maybe you could explain what the concern is, and we can start from there?

sean1588: ahh nvm. yeah I see the tests now. ok that's all I wanted to confirm it validates the structure. thank you!!

sean1588 (approved):

PR #398, opened 2022-10-08
Fix C# project filenames

cnunciato:
When I made these templates, I didn't realize the filenames of C# project names were also templated 🙀 -- just happened to notice that in another template this morning. This fixes 'em up.


scottslowe (approved):
LGTM (and good to know this for future C# work I need to do)

PR #401, opened 2022-10-11
Add container-azure-* templates

cnunciato:
This change adds new templates for `container-azure-{typescript,python,go,csharp}`. Apps return JSON responses using common language-specific frameworks for easy extensibility. Excludes YAML for now, as `docker:Image` is not yet supported in YAML. (See https://github.com/pulumi/pulumi-docker/issues/132.)

Fixes https://github.com/pulumi/devrel-team/issues/286.


scottslowe (approved):
Hey @cnunciato, the C# and Go templates work perfectly but there is a "corner case" that causes Python and TypeScript to fail (related to Docker networking). I don't know that it's necessarily worth trying to address that (I suspect it would be rather rare) so I'll go ahead and approve the PR.

cnunciato:
@scottslowe I wasn't able to get the emoji rendering to work reliably in all browsers (even after ensuring all app servers specified `charset=utf-8`), so I went ahead and removed them. 😢

scottslowe:
I was able to get them to render in Firefox, Chrome, and Safari (after changing the charset). Which browser didn't work for you?

cnunciato:
Go and TypeScript worked fine in all browsers, but C# and Python didn't work in either Chrome or Safari.

PR #380, opened 2022-09-27
feat: base networking for aws

rawkode:
My `pulumi convert` isn't working with AWSx. @scottslowe said he'd try to replicate.

PR #403, opened 2022-10-11
Add templates for base networking on AWS

scottslowe:
This PR adds templates for base networking on AWS in Go, Python, TypeScript, and YAML. C# and Java templates will be added later via a separate PR.

This PR replaces #380.


scottslowe:
Closing until we are ready to work on the base networking templates

PR #404, opened 2022-10-12
Fix serverless-azure-csharp template

cnunciato:
I noticed yesterday that this template wasn't working because the ZIP file that gets created and uploaded to Azure didn't have the internal layout that Azure was expecting. (The archive produced by `Command.Local.Run` lets you filter the files that end up in the ZIP, but it doesn't expose a way to specify a root path for those files, so everything ends up prefixed by the path from which the Command was run, and it doesn't look like there's a way to configure this at the function-definition level, either.) So this change specifies an explicit build path (`publish`, which is also `.gitignore`d by default) and points a `FileArchive` resource at that path instead.


scottslowe (approved):
Approved per our Slack discussion

PR #411, opened 2022-10-18
Add Kubernetes C# templates

cnunciato:
Adds `kubernetes-azure-csharp` and `kubernetes-aws-csharp` to the mix, and adds template metadata for both.

Fixes https://github.com/pulumi/devrel-team/issues/416
Fixes https://github.com/pulumi/devrel-team/issues/421

https://github.com/pulumi/pulumi-hugo/pull/2060 exposes these on landing pages.


scottslowe (approved):
I can't speak to how idiomatic the code is, but the templates and the code work flawlessly. Approved!

PR #412, opened 2022-10-19
Publish metadata.json to dist dir

sean1588:
This PR removes the release publish from the gh workflow and instead published the metadata file to a dist dir. The reason for this is, the way things currently are with the metadata.json being attached as a release artifact we cannot fetch it cross-origin from the pulumi console ui, but we can do it if we add it as a file to the repo and access the raw code form, e.g. [raw.githubusercontent.com/pulumi/templates/](http://raw.githubusercontent.com/pulumi/templates/).... So this PR enables that.

In order to test this, I forked the repo into my personal gh account and tested merging a PR there. Upon doing so the bot updated the file and committed.

image

This PR does the following:
- removes the release creation from .github/workflows/build-metadata.yml
- changes it to build the metadata file and commit it to the dist/ directory when a PR is merged
- add dist dir with current metadata.json
- updates scripts to point to dist dir


.github/workflows/build-metadata.yml
```diff
@@ -8,23 +8,22 @@ jobs:
...
+      - name: Publish metadata.json file
+        uses: stefanzweifel/git-auto-commit-action@v4
         with:
-          repo_token: "${{ secrets.GITHUB_TOKEN }}"
-          automatic_release_tag: "latest"
-          prerelease: true
-          title: "Metadata"
-          files: |
-            metadata.json
+          file_pattern: dist/metadata.json
```
cnunciato, line 28: Would you be up for having this live at `./metadata/dist` instead? Just so it's more closely associated with the content it represents.

cnunciato (commented):
This looks good, one request as to the placement of the file -- let me know what you think.

.github/workflows/build-metadata.yml
sean1588: sounds good. I made the update.

.github/workflows/build-metadata.yml
```diff
@@ -8,23 +8,22 @@ jobs:
...
+        uses: stefanzweifel/git-auto-commit-action@v4
         with:
-          repo_token: "${{ secrets.GITHUB_TOKEN }}"
-          automatic_release_tag: "latest"
-          prerelease: true
-          title: "Metadata"
-          files: |
-            metadata.json
+          file_pattern: metadata/dist/metadata.json
+          commit_message: update metadata.json
```
cnunciato, line 29: Nit (sorry 😇):

```suggestion
          commit_message: Update metadata.json
```

cnunciato (approved):
One tiny nit, otherwise LGTM. I've updated the branch settings to allow for this as well, so we should be good to go.

sean1588:
thanks @cnunciato!

scottslowe (approved):
Seems reasonable to me, approved

PR #413, opened 2022-10-20
Add container-gcp-* templates

cnunciato:
This change adds new templates for `container-gcp-{typescript,python,go,csharp}`. Apps return JSON responses using common language-specific frameworks. Excludes YAML for now, as `docker:Image` is not yet supported.

Fixes https://github.com/pulumi/devrel-team/issues/173.


cnunciato:
Question for reviewers: Should these templates create a new container registry, or use the registry that's enabled on the project? Something tells me we should create a new one, but I'm not finding many examples that do that (most just push to `gcr.io/${project}/${image_name}`. Thoughts, suggestions?

scottslowe:
@cnunciato What is the behavior of the corresponding templates for the other providers? I would try to match that behavior here for consistency among templates across providers (where possible).

cnunciato:
@scottslowe Near as I can tell, there's not really a way to create multiple registries per project; from the docs, it looks like you can push to up to four location-based hosts, but that's it: https://cloud.google.com/container-registry/docs/overview

So I think what we have here should suffice for purpose of the template -- but would love to hear whether @timbohiatt has any additional guidance here for us.

cnunciato:
Merging on @scottslowe's in Slack.

timbohiatt:
Hey @cnunciato @scottslowe thanks for summoning me to another great PR; 

I see this is now merge complete. Sorry about the delay. However for what it is worth here are a few thoughts. That might help for further Example and Template developments. 

- @cnunciato is correct. In a single Google Cloud project you can only have 1 "[Container Registry](https://cloud.google.com/container-registry/docs)". 

- A [Container Registry](https://cloud.google.com/container-registry/docs) is segregated primarily by project/container-image-name:tag and is limited to Docker Only Images.

- [Artifact Registry](https://cloud.google.com/artifact-registry/docs) has additional features to container registry including the ability to create multiple "[Repositories](https://cloud.google.com/artifact-registry/docs/repositories)" each can be bound to a Google Cloud Region for compliance reasons. In addition Artifact registry can house Maven Packages, Go Packages and NPM Packages and a few others, in addition of course to Docker Images. 

My personal suggestion would be all examples or templates later on that require the creation of a Docker Image registry housed on Google Cloud simply use [Artifact Registry](https://cloud.google.com/artifact-registry/docs). I think it would save a headache late on. Plus the examples of creating a Artifact Registry Repository for Docker Images could easily be replicated for Maven, YUM, NPM ect should those examples be required later on. 

Additionally the URL structures for Container Registry and Artifact Registry are quite different. gcr.io vs us-docker.pkg.dev where "us" is the region and "docker" is the registry type. 

Also, I am not sure how this stacks up with other Cloud Providers however at this time Cloud Run only supports Container Images that come from within Container Registry and or Artifact Registry. 

One final consideration from my point. As previously mentioned the Creation of either Artifact Registry or Container registry both require the enablement of Google Cloud API for use with IaC or from within Cloud Console. 

Transitioning [Container Registry](https://cloud.google.com/container-registry/docs) to [Artifact Registry](https://cloud.google.com/artifact-registry/docs) Public Docs:
https://cloud.google.com/artifact-registry/docs/transition/transition-from-gcr
TLDR; Container Registry will now only receive critical security updates and no new features.

Whole lot of information there. Sorry it's late I don't think for this particular PR it would change any outcomes. However I think its context is helpful for potentially future content.

PR #414, opened 2022-10-20
Fix non-templated TypeScript project names

cnunciato:
Happened to notice these need to be templated also. (I'll also follow up with an issue to add some linting for this sort of thing!)

### DIFF
--- a/serverless-gcp-typescript/app/package.json
+++ b/serverless-gcp-typescript/app/package.json
@@ -1,7 +1,6 @@
 {
-    "name": "serverless-gcp-nodejs",
-    "version": "0.0.1",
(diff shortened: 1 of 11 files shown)


scottslowe (approved):
LGTM

serverless-gcp-typescript/app/package.json
```diff
@@ -1,5 +1,5 @@
 {
-    "name": "serverless-gcp-nodejs",
+    "name": "${PROJECT}",
```
cnunciato, line 2: Actually -- this one should be omitted (since it belongs to the container app)! I'll remove.

PR #420, opened 2022-10-22
Add comma when reading eksNodeInstanceType config value

jmsarn:
There is a missing comma when reading the `eksNodeInstanceType` config value causing it to always evaluate to `None` since the key is getting implicitly concatenated to `eksNodeInstanceTypet3.medium`, which doesn't exist in the configuration. Adding the comma will cause the value to read in from the configuration.


cnunciato:
Thanks, @jmsarn!

cnunciato (approved):

PR #424, opened 2022-10-26
remove metadata file

sean1588:


sean1588:
closing this out since we decided this is not what needs to be done and the fix needs to be addressed in the pulumi cli.

PR #427, opened 2022-10-28
Add GCP VM templates

cnunciato:
Adds the following new templates:
- [x] `vm-gcp-yaml` 
- [x] `vm-gcp-typescript`
- [x] `vm-gcp-python`
- [x] `vm-gcp-go`
- [x] `vm-gcp-csharp`

Fixes https://github.com/pulumi/devrel-team/issues/449.


vm-gcp-go/go.mod
```diff
@@ -0,0 +1,65 @@
+module vm-gcp-go
+
+go 1.17
```
scottslowe, line 3: Levi recently recommended to me on one of my templates to increment this to Go 1.18.

vm-gcp-go/main.go
```diff
@@ -0,0 +1,132 @@
...
+	"github.com/pulumi/pulumi/sdk/v3/go/pulumi"
+	"github.com/pulumi/pulumi/sdk/v3/go/pulumi/config"
+)
+
+func main() {
+	pulumi.Run(func(ctx *pulumi.Context) error {
+
+		// Import the program's configuration settings.
+		cfg := config.New(ctx, "")
+		machineType := "f1-micro"
```
scottslowe, line 16: A more idiomatic way of doing it would be to use `config.Try`, then check for an error and set the value. See some of my Go templates for an example (or let me know if you need more detail).

vm-gcp-typescript/package.json
```diff
@@ -0,0 +1,11 @@
+{
```
scottslowe, line 1: Looks like maybe this was generated with 4 spaces instead of 2?

vm-gcp-typescript/tsconfig.json
```diff
@@ -0,0 +1,18 @@
+{
```
scottslowe, line 1: Same comment regarding spaces

scottslowe (changes requested):
Looks great overall, and all the templates worked as expected. Only minor changes suggested.

vm-gcp-go/go.mod
cnunciato: Ah nice, thanks. Will do.

vm-gcp-go/main.go
cnunciato: Thanks, will do!

vm-gcp-typescript/package.json
cnunciato: These generally _are_ four spaces. Only YAML is two-space, generally.

scottslowe (approved):
LGTM

PR #429, opened 2022-10-28
Add Azure VM templates

cnunciato:
Adds the following new templates
- [x] `vm-azure-yaml` 
- [x] `vm-azure-typescript`
- [x] `vm-azure-csharp`
- [x] `vm-azure-python`
- [x] `vm-azure-go`

Fixes https://github.com/pulumi/devrel-team/issues/450.


vm-azure-csharp/.vscode/launch.json
```diff
@@ -0,0 +1,26 @@
+{
```
scottslowe, line 1: Should the entire `.vscode` directory be included in this PR? It would think this is something we want to exclude.

scottslowe (changes requested):
Just the one question/comment about whether we should be including `.vscode` in the commits for this PR. Otherwise, looks great!

vm-azure-csharp/.vscode/launch.json
cnunciato: Whoa! no, it should be excluded. Good catch. I'll also add to .gitignore.

scottslowe (approved):
LGTM

PR #435, opened 2022-11-01
Fix Go module names

cnunciato:
These should be `${PROJECT}` as well, like the others. Includes a similar tweak to the module name of the GCP serverless function template's module name, which is slightly different because of GCP naming restrictions.

### DIFF
--- a/container-azure-go/go.mod
+++ b/container-azure-go/go.mod
@@ -1,4 +1,4 @@
-module azure-container-go
+module ${PROJECT}
 
 go 1.17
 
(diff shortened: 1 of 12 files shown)


scottslowe (approved):
No need to block on this, but I wonder if we should go ahead and update the Go version from 1.17 to 1.18 while we are at it? Otherwise, LGTM.

cnunciato:
> No need to block on this, but I wonder if we should go ahead and update the Go version from 1.17 to 1.18 while we are at it?

Yeah, I paused on this also, but figured I'd keep this PR focused and that we'd circle back and update :alltheversions: in another PR.

PR #436, opened 2022-11-01
Fix exported origin URL

cnunciato:
This is incorrect -- it should be the URL emitted by the storage account, as in the other languages of this template. (The endpoint actually isn't available over HTTP.) [Internal convo here](https://pulumi.slack.com/archives/C043485KTDW/p1667275877490969).


scottslowe (approved):
I haven't tested this, but it seems to bring the TypeScript template back in line with what I'm seeing for the other languages. Approved!

PR #437, opened 2022-11-01
Add a test that ensures arch templates also exist

cnunciato:
The `metadata.json` file has an `architectures` section that lists the templates that belong to each architecture. Because we manage these lists by hand (by appending template names as strings to each template group), it's technically possible to list a template that doesn't actually exist, which could break consumers of this file (e.g., the Service) that rely on it for creating new projects. So this change adds a couple of tests to enure every template referenced in `metadata.json` also exists in the repository.

### DIFF
--- a/tests/metadata/metadata.test.js
+++ b/tests/metadata/metadata.test.js
@@ -83,5 +83,11 @@ describe("Template metadata", () => {
             expect(templates["gcp-yaml"]).toBeDefined();
             expect(templates["azure-java"]).toBeDefined();
         });
+
+        it("only contains templates that exist", () => {
+            Object.keys(templates).forEach(template => {
+                expect(() => fs.readdirSync(`${template}`)).not.toThrow();
+            });
+        });
     });
 });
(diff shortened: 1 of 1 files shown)


tests/metadata/metadata.test.js
```diff
@@ -71,5 +83,11 @@ describe("Template metadata", () => {
             expect(templates["gcp-yaml"]).toBeDefined();
             expect(templates["azure-java"]).toBeDefined();
         });
+
+        it("only contains templates that exist in the repository", () => {
```
cnunciato, line 87: Figured we might as well include the full list in these checks as well.

scottslowe (approved):
Looks reasonable (I haven't tested this)

cnunciato:
@scottslowe I made sure this worked, yeah. You can run these tests locally either with `make metadata` (which builds and then tests the metadata file -- this is what runs in CI) or just `yarn && yarn test` if you want to see what happens after modifying `metadata.json` directly.

PR #438, opened 2022-11-01
Fix Azure/Google Cloud Go-Container templates

nimbinatus:
The current template fails on M1 Macs due to the architecture differences. This change sidesteps the issue with env variables in Go with docker.Image.


nimbinatus:
Note that I'm not a huge fan of the hardcoding, but this might be the only thing to do right now until buildx lands.

scottslowe (approved):
LGTM

PR #445, opened 2022-11-03
Drop cloud references from template group names

cnunciato:
These group names are more verbose than what's needed in the Service UI, so this change strips out the cloud names to make them a bit tidier. Additional context [in this thread](https://pulumi.slack.com/archives/C0289AASSG4/p1667500236565769).

This change will have no user-facing impact until the accompanying Service change (https://github.com/pulumi/pulumi-service/pull/11034) is merged.

Part of https://github.com/pulumi/pulumi-service/issues/11031.

### DIFF
--- a/metadata/dist/metadata.json
+++ b/metadata/dist/metadata.json
@@ -5,7 +5,7 @@
             "slug": "static-website",
             "groups": [
                 {
-                    "name": "AWS Static Website",
+                    "name": "Static Website",
                     "kind": "architecture",
                     "parent": "static-website",
                     "slug": "aws",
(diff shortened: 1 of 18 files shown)


scottslowe:
Quick Q: Will "Web Application" _only_ show up if Kubernetes is selected as the platform?

cnunciato:
I defer to @sean1588, but I'm pretty sure the answer to that is yes (because its `cloud` is `kubernetes`).

scottslowe (approved):
Approved assuming the behavior is as discussed in the comments (otherwise "Web Application" won't convey to the user enough information).

sean1588:
> I defer to @sean1588, but I'm pretty sure the answer to that is yes (because its cloud is kubernetes).

yes it should. Let me pull this branch down locally and gen the metadata file and test it in the console to verify.

sean1588:
ok, just verified. yes those options only show up when selecting kubernetes. this is how it looks now.
image

sean1588 (approved):
:shipit:

PR #454, opened 2022-11-11
Update examples to use config instead of configuration

### LINKED ISSUE #453: Update examples to use `config` instead of `configuration`
`pulumi convert` requires a bit more refactoring before `config` can be safely used and `configuration` can be fully deprecated.

aq17:
pulumi-yaml `configuration` is being deprecated in favor of the project-level `config` block 
Fixes #453

### DIFF
--- a/container-aws-yaml/Pulumi.yaml
+++ b/container-aws-yaml/Pulumi.yaml
@@ -17,15 +17,15 @@ template:
       description: The amount of memory to allocate for the container
       default: 128
 
-configuration:
+config:
   containerPort:
-    type: Number
+    type: integer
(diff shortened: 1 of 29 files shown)


iwahbe (changes requested):
The `type:` field needs to be changed.`Number` should be `integer` if its an integer. `String` should be `string`. There is currently no way to have a float AFAIK.

aq17:
@iwahbe I think this can be merged – previous PR's also are failing on these tests

iwahbe (approved):
If this PR is getting the same failures as other master, then I agree we can merge this.

PR #455, opened 2022-11-16
Skip benchmarks to save disk space on container

t0yv0:
https://github.com/pulumi/templates/actions/runs/3475152250/jobs/5809065821 an example of running out of disk space. Skipping these for now should speed up the job and save disk space.


RobbieMcKinstry (approved):

t0yv0:
These time series are not in the dashboard yet afaik but yes if they flowed into the Redshift table they will stop now. I think that's a fine tradeoff. If we wanted that data we'd probably need to parallelize/distribute this job across N containers.

RobbieMcKinstry:
Yes, that makes sense. IMO it's fine if it stops the data flowing into Redshift but doesn't affect the dash. Since they're not in the dash, this is 👍🏻

PR #467, opened 2022-11-28
Update examples to use config instead of configuration

aq17:


aq17:
`kubernetes-azure-yaml` has not been re-updated yet since it was the template causing failures due to project config being unable to define provider config options

iwahbe (approved):
We've done a revert and reapply already. What can we do to make sure this doesn't happen again? For Pulumi YAML at the least, can we run all of these templates against mock providers to make sure the files work?

aq17:
> We've done a revert and reapply already. What can we do to make sure this doesn't happen again? For Pulumi YAML at the least, can we run all of these templates against mock providers to make sure the files work?

I tested each modified template with `pulumi preview` successfully. I think the relaxed type check fix between `number` and `int` covers the main diff between `configuration` -> `config`

PR #468, opened 2022-11-29
Target new version of AWSX in new AWS templates

pierskarsenbarg:


danielrbradley (approved):
Great ta ... it's a little odd that we don't actually use AWSx in the template - only add the reference but that's fine

PR #470, opened 2022-11-30
Unpin .NET dependencies

### LINKED ISSUE #469: serverless-aws-typescript and -csharp templates fail to compile
Likely related to dependency updates (https://github.com/pulumi/templates/pull/466).

cnunciato:
Allows programs to pick up minor and patch versions of Pulumi dependencies. (One of these templates, `serverless-aws-csharp`, is currently failing  because it's pinned to `Aws = 5.13.0`, but `AwsApiGateway` is trying to upgrade it to `>= 5.21.1`.)

```
...
my-serverless-app.csproj : error NU1605: Detected package downgrade: Pulumi.Aws from 5.21.1 to 5.13.0. Reference the package directly from the project to select a different version.
```
 
Fixes #469.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,5 @@ go.sum
 Pulumi.*.yaml
 node_modules
 .vscode
+[Bb]in/
+[Oo]bj/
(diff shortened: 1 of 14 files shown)


scottslowe (approved):
I didn't test these changes, but they all seem reasonable to me. However, I did note that `container-gcp-csharp/${PROJECT}.csproj` has a minor version for the Docker provider, and `serverless-gcp-csharp/${PROJECT}.csproj` has the full version for SyncedFolder (while others don't). I don't know if you want to correct those with an additional commit or not.

cnunciato:
@scottslowe Good catches, thanks! I'll fix those up as well (missed 'em).

PR #478, opened 2022-12-09
Fix YAML template for Kubernetes on GCP

### LINKED ISSUE #477: Provider-namespace configuration values should not have a type
### What happened?

Architecture templates that leverage provider-namespace configuration values (like `aws:region` or `gcp:project`) are broken following the implementation of hierarchical configuration (as implemented via https://github.com/pulumi/pulumi/pull/10832). This appears to be due to the fact that the architecture templates included a `type` field for provider-namespace configuration values, and that field is no longer supported/allowed.

### Steps to reproduce

1. Run `pulumi new kubernetes-gcp-yaml`.
2. Observe the error that results:
```
error: template 'kubernetes-gcp-yaml' is currently broken: could not unmarshal '/Users/slowe/.pulumi/templates/kubernetes-gcp-yaml/Pulumi.yaml': Configuration key 'gcp:project' is not namespaced by the project and should not define a type
```

### Expected Behavior

It's expected that `pulumi new` will work and create a new project as expected.

### Actual Behavior

The `pulumi new` command fails with an error, and there is no obvious workaround for users.

### Output of `pulumi about`

```
CLI
Version      3.49.0
Go Version   go1.19.4
Go Compiler  gc

Host
OS       darwin
Version  12.6.1
Arch     arm64

Backend
Name           pulumi.com
URL            https://app.pulumi.com/scottslowe
User           scottslowe
Organizations  scottslowe

Pulumi locates its logs in /var/folders/c_/1pdwtg4j5zx4dvx5t6qp9v180000gn/T/ by default
```

### Additional context

Confirmed on two different macOS systems, one running Pulumi 3.49.0 (that's the output of `pulumi about` shown above) and another system running Pulumi 3.45.0.

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

scottslowe:
This PR fixes issues with the YAML template for Kubernetes on GCP, which was affected by the issue described in #477. No other templates were identified as being affected by #477.

This PR fixes #477.

### DIFF
--- a/kubernetes-gcp-yaml/Pulumi.yaml
+++ b/kubernetes-gcp-yaml/Pulumi.yaml
@@ -14,10 +14,6 @@ template:
       description: The desired number of nodes PER ZONE in the nodepool
 
 config:
-  gcp:project:
-    value: changeme
-  gcp:region:
-    value: us-central1
   nodesPerZone:
     type: integer
     default: 1
--- a/kubernetes-gcp-yaml/Pulumi.yaml
+++ b/kubernetes-gcp-yaml/Pulumi.yaml
@@ -15,10 +15,9 @@ template:
 
 config:
   gcp:project:
-    type: string
+    value: changeme
   gcp:region:
-    type: string
-    default: us-central1
+    value: us-central1
   nodesPerZone:
     type: integer
     default: 1
--- a/kubernetes-gcp-yaml/Pulumi.yaml.append
+++ b/kubernetes-gcp-yaml/Pulumi.yaml.append
@@ -1,8 +1,4 @@
 config:
-  gcp:project:
-    value: changeme
-  gcp:region:
-    value: us-central1
   nodesPerZone:
     type: integer
     default: 1
--- a/kubernetes-gcp-yaml/Pulumi.yaml.append
+++ b/kubernetes-gcp-yaml/Pulumi.yaml.append
@@ -1,9 +1,8 @@
 config:
   gcp:project:
-    type: string
+    value: changeme
   gcp:region:
-    type: string
-    default: us-central1
+    value: us-central1
   nodesPerZone:
     type: integer
     default: 1


kubernetes-gcp-yaml/Pulumi.yaml
```diff
@@ -15,10 +15,9 @@ template:
 
 config:
   gcp:project:
-    type: string
+    value: changeme
   gcp:region:
-    type: string
-    default: us-central1
+    value: us-central1
```
cnunciato, line 20: I _think_ the fix for this is to remove these provider config settings from the root `config` block, as they'll be set during `pulumi new` using defaults from `template.config`. When I run `pulumi new` with this branch, for example, I don't see `changeme` in the prompt as a default, and after creation I end up with an empty string in `Pulumi.dev.yaml`: 

```yaml
config:
  asasdasdfa:nodesPerZone: "1"
  gcp:project: ""
  gcp:region: us-central1
```

Tried with 3.48 and 3.49 and got the same result. Looking at other `*-gcp-yaml` templates, it seems they all leave these it out of the program config sections, so we should be good just removing it.

cnunciato (changes requested):

kubernetes-gcp-yaml/Pulumi.yaml
scottslowe: We can't leave those out; they are referenced later in the template. In order to be able to reference them later, they have to be included in the `config` block. And there has to be at least a value key under each of them, because the template won't work otherwise.

cnunciato: I see, ok. We’ll if this fixes it, let’s go ahead and merge — we can always come back it. Thanks for the explanation and the fix!

cnunciato (approved):

kubernetes-gcp-yaml/Pulumi.yaml
lukehoban: cc @aq17 @AaronFriel Curious what we would expect to be used here?  It doesn't feel right to require that we include dummy values like this?

aq17: Yes, for keys that should have a default, using `value` instead of `type` and `default` is correct, but using a dummy seems like it could lead to confusion. 
We could consider allowing project config keys to just be declared and not require a value within the `Pulumi.yaml` file?

aq17: Actually, it is possible to omit `gcp:project` and `gcp:region` from the template and still reference them; when running `pulumi up`, you should get 
```
    warning: unable to detect a global setting for GCP Project;
    Pulumi will rely on per-resource settings for this operation.
    Set the GCP Project by using:
    	`pulumi config set gcp:project <project>` 
```
if it isn't set in your stack config, which seems preferable to using dummy values.

scottslowe: @aq17 Interesting; I was under the impression that the only way to use a configuration value (provider-namespaced or otherwise) was to include it in the `config` block (the template does reference `gcp:project` later on). I'll try removing it from the `config` block and report back.

cnunciato: > I think the fix for this is to remove these provider config settings from the root config block

> Actually, it is possible to omit gcp:project and gcp:region from the template and still reference them

Finally got a chance to try this locally myself, and just removing those two settings does seem to do the trick, yeah.

We should also remove them from the template's page on pulumi.com: https://www.pulumi.com/templates/kubernetes/gcp/#customizing-the-project

scottslowe: I also tested this locally and it seemed to work; I wanted to run one more set of tests to be sure. Agreed that if we remove these settings we need to update the landing page as well.

scottslowe: I've run a few tests and can't find any ill effects from removing `gcp:project` and `gcp:region` from the top-level `config` block, so I'll push another commit that just removes those and re-request reviews before merging.

PR #483, opened 2022-12-14
Bump deps for static-website and serverless templates

### LINKED ISSUE #482: Update templates to use latest Synced Folder
Follow-up to #430, we need to update any templates that use synced-folder to use its latest version.

Should also update any peer deps that use older versions of AWS, Azure, or GCP as well, so we aren't pulling multiple versions of provider plugins.

cnunciato:
Upgrades dependencies on the `serverless-*` and `static-website-*` templates. 

I tested all of these manually as I went through them, so there shouldn't be any issues, but a few things to note:

* I noticed the API Gateway package was pinned to an older version of `pulumi-aws` (causing templates to pull down multiple versions of it), so I submitted a PR to fix that here: https://github.com/pulumi/pulumi-aws-apigateway/pull/70
* With these updates, the `serverless-azure-go` template emits a deprecation notice that I haven't tracked down yet but that appears to have something to do with the building of the packaged Go app. Doesn't interfere with deployment at all; is just a bit noisy. I'll see if I can figure out where it's coming from and address if possible in a follow-up.
* A new GCP shipped just as synced-folder was updated yesterday, so in some cases, we're still downloading two versions of `pulumi-gcp`. Would be awesome if we could let these transitive deps drift a little to reduce the download burden, but I'm not sure that's possible with component packages yet.

Fixes #482.

### DIFF
--- a/serverless-aws-csharp/${PROJECT}.csproj
+++ b/serverless-aws-csharp/${PROJECT}.csproj
@@ -9,7 +9,7 @@
 	<ItemGroup>
 		<PackageReference Include="Pulumi" Version="3.*" />
 		<PackageReference Include="Pulumi.Aws" Version="5.*" />
-		<PackageReference Include="Pulumi.AwsApiGateway" Version="*" />
+		<PackageReference Include="Pulumi.AwsApiGateway" Version="1.*" />
 	</ItemGroup>
 
 </Project>
--- a/serverless-aws-go/go.mod
+++ b/serverless-aws-go/go.mod
@@ -3,16 +3,23 @@ module ${PROJECT}
 go 1.17
 
 require (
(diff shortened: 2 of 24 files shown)


scottslowe:
I noticed in the TypeScript templates that you switched `pulumi/pulumi` from `^3.0.0` to `^3.49.0`. Is pinning the minor version needed for some reason?

cnunciato:
> I noticed in the TypeScript templates that you switched `pulumi/pulumi` from `^3.0.0` to `^3.49.0`. Is pinning the minor version needed for some reason?

Yeah, it’s just the result of running `yarn add`; the end result is the same (installs will use the latest minor version either way), I just opted to accept what the tool produced rather than fiddle with it manually.

Issue #481, opened 2022-12-13
github-go dependency problem


figadore:
### What happened?

Creates an invalid set of dependencies

go.mod
```
go 1.17

require (
        github.com/pulumi/pulumi-github/sdk/v4 v5.0.0
        github.com/pulumi/pulumi/sdk/v3 v3.44.2
)

```

main.go
```
package main

import (
        "github.com/pulumi/pulumi-github/sdk/v4/go/github"
        "github.com/pulumi/pulumi/sdk/v3/go/pulumi"
)
```

### Steps to reproduce

`pulumi new github-go`

### Expected Behavior

initialize a pulumi stack

### Actual Behavior

```
...
Saved config                                                                                                                                                                                                                                
                                                                                                                                                                                                                                            
Installing dependencies...                                                                                                                                                                                                                  
                                                                                                                                                                                                                                            
go: errors parsing go.mod:                                                                                                                                                                                                                  
/some/path/go.mod:6:2: require github.com/pulumi/pulumi-github/sdk/v4: version "v5.0.0" invalid: should be v4, not v5                                                                                        
error: installing dependencies failed; rerun manually to try again, then run `pulumi up` to perform an initial deployment: `go mod tidy` failed to install dependencies: exit status 1      
```

### Output of `pulumi about`

```
CLI          
Version      3.49.0
Go Version   go1.19.3
Go Compiler  gc

Plugins
NAME  VERSION
go    unknown

Host     
OS       debian
Version  bookworm/sid
Arch     x86_64

This project is written in go: executable='/usr/local/go/bin/go' version='go version go1.19.2 linux/amd64'

Backend        
Name           pop-os
URL            file://~
User           reese
Organizations  

Pulumi locates its logs in /tmp by default
warning: Failed to get information about the Pulumi program's dependencies: failed to get modules: exit status 1
warning: Failed to get information about the current stack: No current snapshot
```

### Additional context

Not sure if this issue belongs here or github.com/pulumi/pulumi-github

I also saw [in the docs](https://www.pulumi.com/docs/get-started/gcp/begin/) that go 1.18 was required, should that be reflected in this template in go.mod?

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

squaremo:
I can confirm this, same thing happens for me. There's something pretty wacky going on with the versioning here (and in pulumi-github#276).

lukehoban:
[...]

PR #494, opened 2023-01-12
Skip updates, {vm,kubernetes}-azure-yaml

cnunciato:
As a follow-up to https://github.com/pulumi/devrel-team/issues/464, this change skips running full updates for all templates, relying instead on successful previews, and adds `vm-azure-yaml` and `kubernetes-azure-yaml` to the list of skipped tests.

Doesn't represent the ideal state for sure, but given the state we're in now, we think this is an acceptable trade-off to get back to reliably passing while we explore alternative ways to run full deployments for some or all templates.

Note that this change also skips testing `github-go` (temporarily) due to https://github.com/pulumi/templates/issues/481. I'll remove it from the list once that issue's fixed.

Fixes https://github.com/pulumi/devrel-team/issues/487.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,7 +19,7 @@ env:
   GOOGLE_PROJECT: pulumi-ci-gcp-provider
   GOOGLE_PROJECT_NUMBER: 895284651812
   LINODE_TOKEN: ${{ secrets.LINODE_TOKEN }}
-  SKIPPED_TESTS: "alicloud,digitalocean,kubernetes,openstack,equinix-metal,civo,aiven,auth0,github,oci,java-jbang,java-gradle,azuredevops"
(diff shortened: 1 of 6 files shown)


.github/workflows/ci.yml
```diff
@@ -125,14 +125,11 @@ jobs:
         env:
           PULUMI_PYTHON_CMD: python
           TESTPARALLELISM: 3
-          SKIPPED_TESTS: "alicloud,digitalocean,kubernetes,openstack,equinix-metal,civo,aiven,auth0,github,oci,java-jbang,java-gradle,azuredevops,container"
```
cnunciato, line 128: Not sure why these are overridden -- I believe the steps should inherit values set on line 22.

scottslowe (approved):
Approved

PR #495, opened 2023-01-12
Use GITHUB_TOKEN in curl requests, run generate-go-mod.sh

cnunciato:
Running this script without a `GITHUB_TOKEN` [occasionally yields `null` values in `go.mod` files](https://github.com/pulumi/templates/pull/493) because of GitHub rate limiting. This change adds the ambient `GITHUB_TOKEN` to `curl` request headers and re-runs the script to regenerate updated values.  

Also "downgrades" `github-go` to `v4` while we sort out what's going on with #481.

### DIFF
--- a/equinix-metal-go/main.go
+++ b/equinix-metal-go/main.go
@@ -1,7 +1,7 @@
 package main
 
 import (
-	metal "github.com/pulumi/pulumi-equinix-metal/sdk/v2/go/equinix"
+	metal "github.com/pulumi/pulumi-equinix-metal/sdk/v3/go/equinix"
 	"github.com/pulumi/pulumi/sdk/v3/go/pulumi"
 )
 
--- a/aws-go/go.mod
+++ b/aws-go/go.mod
@@ -3,6 +3,6 @@ module ${PROJECT}
 go 1.17
 
 require (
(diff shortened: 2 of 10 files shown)


scottslowe (approved):
LGTM. At some point (not as part of this PR), we need to think about updating the Go version for all these templates.

equinix-metal-go/main.go
```diff
@@ -1,7 +1,7 @@
 package main
 
 import (
-	metal "github.com/pulumi/pulumi-equinix-metal/sdk/v2/go/equinix"
+	metal "github.com/pulumi/pulumi-equinix-metal/sdk/v3/go/equinix"
```
cnunciato, line 4: This line needs updated as well (to correspond with the updated version).

github-go/main.go
```diff
@@ -1,7 +1,7 @@
 package main
 
 import (
-	"github.com/pulumi/pulumi-github/sdk/v4/go/github"
+	"github.com/pulumi/pulumi-github/sdk/v5/go/github"
```
cnunciato, line 4: Similarly here.

scottslowe (approved):
LGTM

Issue #448, opened 2022-11-04
Fix template description for Kubernetes on GCP templates


scottslowe:
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details

It looks like we missed one template in the recent PR (#447) to make sure the template descriptions are consistent. (See Luke's comment on that PR.)

PR #496, opened 2023-01-12
Fix Kubernetes template description

cnunciato:
Fixes https://github.com/pulumi/templates/issues/448.


scottslowe (approved):
LGTM, thanks for catching these that slipped past me!

PR #499, opened 2023-01-13
Bump go versions to 1.18

cnunciato:
Brings to [one version behind current](https://go.dev/doc/devel/release#policy).

Fixes https://github.com/pulumi/templates/issues/508.


Frassle (approved):

scottslowe (approved):
Changes seem reasonable and tests seem to be passing, so approved

PR #500, opened 2023-01-13
Fix environment-variable reference

cnunciato:
Our PR tests use an environment variable, `PULUMI_TEMPLATE_LOCATION`, to pass the path of the checked-out branch to `pulumi new` to ensure that the tests run against the content of the PR. This environment variable, however, hasn't been getting set for some time, so the tests have actually been running against the templates in pulumi/templates@master, using  the default behavior of `pulumi new` (which clones over https). This change fixes that.

Interestingly, this change also seems to have surfaced new failures on Windows; specifically, all of the `fsharp` templates are now failing on the Windows runners, and a couple of `visualbasic` ones are now, too. This appears to have something to do with a UTF-8 byte-order mark that's apparently set by Visual Studio on Windows, but removing those marks doesn't seem to fix it, so for now, since I've verified these templates actually do work on Windows, and the issue seems specific to running `pulumi new` with a file path, I've commented them out as well, and will file an issue to investigate further.

### DIFF
--- a/.github/workflows/run-templates-command.yml
+++ b/.github/workflows/run-templates-command.yml
@@ -17,7 +17,7 @@ env:
   GOOGLE_PROJECT: pulumi-ci-gcp-provider
   GOOGLE_PROJECT_NUMBER: 895284651812
   LINODE_TOKEN: ${{ secrets.LINODE_TOKEN }}
-  SKIPPED_TESTS: "alicloud,digitalocean,kubernetes,openstack,equinix-metal,civo,aiven,auth0,github,oci,java-jbang,java-gradle,azuredevops,container,vm-azure-yaml,kubernetes-azure-yaml,github-go"
+  SKIPPED_TESTS: "alicloud,digitalocean,kubernetes,openstack,equinix-metal,civo,aiven,auth0,github,oci,java-jbang,java-gradle,azuredevops,container,vm-azure-yaml"
   PULUMI_API: https://api.pulumi-staging.io
   ARM_CLIENT_ID: ${{ secrets.ARM_CLIENT_ID }}
   ARM_CLIENT_SECRET: ${{ secrets.ARM_CLIENT_SECRET }}
@@ -25,6 +25,7 @@ env:
   ARM_TENANT_ID: ${{ secrets.ARM_TENANT_ID }}
   AZURE_LOCATION: westus
   TESTPARALLELISM: 10
+  PULUMI_TEMPLATE_LOCATION: ${{ github.workspace}}
 jobs:
   sentinel:
     if: github.event_name == 'repository_dispatch' ||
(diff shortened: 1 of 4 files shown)


.github/workflows/run-templates-command.yml
```diff
@@ -17,7 +17,7 @@ env:
   GOOGLE_PROJECT: pulumi-ci-gcp-provider
   GOOGLE_PROJECT_NUMBER: 895284651812
   LINODE_TOKEN: ${{ secrets.LINODE_TOKEN }}
-  SKIPPED_TESTS: "alicloud,digitalocean,kubernetes,openstack,equinix-metal,civo,aiven,auth0,github,oci,java-jbang,java-gradle,azuredevops,container,vm-azure-yaml,kubernetes-azure-yaml,github-go"
+  SKIPPED_TESTS: "alicloud,digitalocean,kubernetes,openstack,equinix-metal,civo,aiven,auth0,github,oci,java-jbang,java-gradle,azuredevops,container,vm-azure-yaml,kubernetes-azure-yaml"
```
cnunciato, line 20: This was actually redundant, as `github` is already in the list.

.github/workflows/run-templates-command.yml
```diff
@@ -104,10 +104,7 @@ jobs:
       - name: Authenticate to Google Cloud
         uses: google-github-actions/auth@v0
         with:
-          workload_identity_provider: projects/${{ env.GOOGLE_PROJECT_NUMBER
-            }}/locations/global/workloadIdentityPools/${{
-            env.GOOGLE_CI_WORKLOAD_IDENTITY_POOL }}/providers/${{
-            env.GOOGLE_CI_WORKLOAD_IDENTITY_PROVIDER }}
+          workload_identity_provider: projects/${{ env.GOOGLE_PROJECT_NUMBER }}/locations/global/workloadIdentityPools/${{ env.GOOGLE_CI_WORKLOAD_IDENTITY_POOL }}/providers/${{ env.GOOGLE_CI_WORKLOAD_IDENTITY_PROVIDER }}
```
cnunciato, line 107: Just for readability.

tests/internal/testutils/templates.go
```diff
@@ -32,7 +32,7 @@ func FindAllTemplates(t *testing.T, templateUrl string) []TemplateInfo {
 		templateName := t.Name
 		templatePath := templateName
 		if templateUrl != "" {
-			templatePath = path.Join(templateUrl, templateName)
+			templatePath = filepath.Join(templateUrl, templateName)
```
cnunciato, line 35: Made this change after noticing the paths on Windows contained a mixture of backward and forward slashes. Using `filepath` instead of `path` fixes this up.

.github/workflows/run-templates-command.yml
```diff
@@ -104,17 +105,12 @@ jobs:
...
           service_account: ${{ env.GOOGLE_CI_SERVICE_ACCOUNT_EMAIL }}
       - name: Setup gcloud auth
         uses: google-github-actions/setup-gcloud@v0
         with:
           install_components: gke-gcloud-auth-plugin
-      - name: Set the Template URL using local repo path
```
cnunciato, line 116: I believe this is old GHA syntax. It might've worked at one time, but definitely does not work today.

.github/workflows/run-templates-command.yml
```diff
@@ -104,17 +105,12 @@ jobs:
...
-        run: echo "PULUMI_TEMPLATE_LOCATION=${{ github.workspace}}" >> $GITHUB_ENV
```
cnunciato, line 117: I believe this is old GHA syntax. It might've worked at one time, but definitely does not work today.

.github/workflows/run-templates-command.yml
```diff
@@ -17,14 +17,15 @@ env:
   GOOGLE_PROJECT: pulumi-ci-gcp-provider
   GOOGLE_PROJECT_NUMBER: 895284651812
   LINODE_TOKEN: ${{ secrets.LINODE_TOKEN }}
-  SKIPPED_TESTS: "alicloud,digitalocean,kubernetes,openstack,equinix-metal,civo,aiven,auth0,github,oci,java-jbang,java-gradle,azuredevops,container,vm-azure-yaml,kubernetes-azure-yaml,github-go"
```
cnunciato, line 20: These last two are redundant, as `kubernetes` and `github` already exist in the list (and they're partial-matched).

scottslowe (approved):
Approved! Thanks for helping us improve the quality of the testing framework for this repository.

cnunciato:
Filed: https://github.com/pulumi/templates/issues/503

PR #504, opened 2023-01-17
Upgrade actions to address warnings

cnunciato:
We've accumulated a number of warnings in our actions results:

image

To address these warnings, this change upgrades everything that can be upgraded and removes the one step we have that uses `set-output`. Couple notes:

* pulumi/actions-install-pulumi-cli is now called [pulumi/setup-pulumi](https://github.com/pulumi/setup-pulumi) and has been upgraded to the latest version, but the latest version unfortunately still runs Node 12. https://github.com/pulumi/setup-pulumi/issues/36 tracks addressing that.
* aws-actions/configure-aws-credentials also still runs on Node 12, so will continue to warn until https://github.com/aws-actions/configure-aws-credentials/issues/489 is addressed.
* The step that uses `set-output` appears to post a comment with a link back to the build job, presumably for external contributors, but since contributors can easily navigate to these build jobs in the usual way (i.e., by clicking the Details links in the Checks result), it's superfluous, so I've chosen to remove rather than refactor it.

Fixes https://github.com/pulumi/templates/issues/509.

### DIFF
--- a/.github/workflows/build-metadata.yml
+++ b/.github/workflows/build-metadata.yml
@@ -9,19 +9,19 @@ jobs:
     name: Build
     steps:
       - name: Install Node
-        uses: actions/setup-node@v1
+        uses: actions/setup-node@v3
         with:
-          node-version: "16.x"
-      
+          node-version: 16.x
+
       - name: Check out branch
         uses: actions/checkout@v3
         with:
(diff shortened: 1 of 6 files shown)


scottslowe (approved):
LGTM

PR #506, opened 2023-01-18
Run `push` jobs against local checkout as well

cnunciato:
Updates the `push` tests to run against the local checkout also (as we do for PRs, as of https://github.com/pulumi/templates/pull/500). Should make the README badge go green. 🤞 

Fixes https://github.com/pulumi/templates/issues/510.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,10 +1,6 @@
 name: Run Template Tests on Push or as Part of a Release
 on:
-
-  # Temporary to test this change.
-  pull_request:
-  # push:
-
+  push:
     branches:
       - master
   workflow_dispatch: {}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,6 +1,10 @@
 name: Run Template Tests on Push or as Part of a Release
 on:
-  push:
+
+  # Temporary to test this change.
+  pull_request:
+  # push:
+
     branches:
       - master
   workflow_dispatch: {}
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -1,6 +1,10 @@
 name: Run Template Tests on Push or as Part of a Release
 on:
-  push:
+
+  # Temporary to test this change.
+  pull_request:
+  # push:
```
cnunciato, line 6: Will undo this before merging.

.github/workflows/ci.yml
```diff
@@ -19,7 +23,7 @@ env:
   GOOGLE_PROJECT: pulumi-ci-gcp-provider
   GOOGLE_PROJECT_NUMBER: 895284651812
   LINODE_TOKEN: ${{ secrets.LINODE_TOKEN }}
-  SKIPPED_TESTS: "alicloud,digitalocean,kubernetes,openstack,equinix-metal,civo,aiven,auth0,github,oci,java-jbang,java-gradle,azuredevops,container,vm-azure-yaml,kubernetes-azure-yaml,github-go"
+  SKIPPED_TESTS: "alicloud,digitalocean,kubernetes,openstack,equinix-metal,civo,aiven,auth0,github,oci,java-jbang,java-gradle,azuredevops,container,vm-azure"
```
cnunciato, line 26: Since `kubernetes` and `github` are already in the list, these can be removed, and we want to skip all `vm-azure` (not just YAML) because they all require `sshPublicKey`.

scottslowe (approved):
Approved

PR #515, opened 2023-01-19
Fix go.mod templating script

cnunciato:
This change makes a few updates:

* Bumps the Go versions in the template files that we use to regenerate `go.mod` files (e.g., in response to a CLI release) to 1.18 (to align with the templates themselves)
* Alphabetizes the names of the providers for which we do this
* Adjusts the commit message and title of the PRs we generate from these events (to make it clearer what these PRs actually are and where they come from)
* Re-runs the script to get the `go.mod`s updated

Replaces #512.


scottslowe (approved):
Approved

PR #516, opened 2023-01-19
Update daily CLI test to use local templates

cnunciato:
In #500 and #506, we updated the tests to run against the local checkout. This change updates the daily CLI tests (which [have been failing in similar ways](https://github.com/pulumi/templates/actions/runs/3956441898/jobs/6775658807#step:21:1408)) to do the same.


scottslowe (approved):
LGTM

PR #520, opened 2023-02-02
Add scala template

asaf400:
I have noticed that while [this blog post](https://www.pulumi.com/blog/announcing-infrastructure-as-code-with-java-and-pulumi/), and [this closed issue](https://github.com/pulumi/pulumi/issues/1539) exist,
in the blog it is stated: 
> Declare infrastructure using programs, classes, and libraries written in Java or other JVM languages (Kotlin, **Scala**, Clojure, Groovy, etc.).

however, `pulumi new` command doesn't have a scala template, specifically I was looking for the aws template,
I have found 2 resources regarding an example:
https://github.com/pulumi/pulumi-java/pull/791
https://blog.ediri.io/how-to-use-pulumi-and-scala-to-deploy-a-minecraft-server-on-a-digitalocean-droplet

with those examples, I was able to recreate the same 'default' aws bucket example,
and `pulumi up` works, please consider adding this to the template list..

I know this is only aws, and it is not a all scala templates for all cloud providers, but,
This is a good start even for other providers, I know myself, I would have started with a Scala-Azure template if it was available, and just change the dependencies + imports..

### DIFF
--- a/aws-scala/src/main/scala/Main.scala
+++ b/aws-scala/src/main/scala/Main.scala
@@ -6,7 +6,11 @@ import com.pulumi.aws.s3.Bucket
 object App {
   def main(args: Array[String]): Unit = {
     Pulumi.run { (ctx: Context) =>
+
+      // Create an AWS resource (S3 Bucket)
       var bucket = new Bucket("my-bucket");
+
+      // Export the name of the bucket
       ctx.`export`("bucketName", bucket.bucket())
     }
   }
--- a/aws-scala/src/main/scala/Main.scala
+++ b/aws-scala/src/main/scala/Main.scala
@@ -1,0 +1,13 @@
+package myproject
+
+import com.pulumi.{Context, Pulumi}
+import com.pulumi.aws.s3.Bucket
+
+object App {
+  def main(args: Array[String]): Unit = {
+    Pulumi.run { (ctx: Context) =>
+      var bucket = new Bucket("my-bucket");
+      ctx.`export`("bucketName", bucket.bucket())
+    }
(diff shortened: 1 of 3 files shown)


cnunciato:
Thanks for the contribution @asaf400! We'll get this looked at ASAP.

aws-scala/src/main/scala/Main.scala
```diff
@@ -0,0 +1,13 @@
...
+
+import com.pulumi.{Context, Pulumi}
+import com.pulumi.aws.s3.Bucket
+
+object App {
+  def main(args: Array[String]): Unit = {
+    Pulumi.run { (ctx: Context) =>
+      var bucket = new Bucket("my-bucket");
+      ctx.`export`("bucketName", bucket.bucket())
+    }
```
cnunciato, line 11: Most of our templates have comments that call out the important bits, so would be good to add them here, too:

```suggestion
    Pulumi.run { (ctx: Context) =>

      // Create an AWS resource (S3 Bucket)
      var bucket = new Bucket("my-bucket");

      // Export the name of the bucket
      ctx.`export`("bucketName", bucket.bucket())
    }
```

cnunciato (commented):
Thanks again! 🙌  To my my non-Scala-expert eyes, this looks great, and worked as expected when I ran through it locally. Only a few minor suggestions for you to consider, and then we should be able to merge it in!

asaf400:
@cnunciato thanks for the suggestions, I accepted both of them

cnunciato (approved):
👏

PR #531, opened 2023-03-09
Conditionally test some templates end to end

### LINKED ISSUE #533: Enable full updates in tests of key templates
In https://github.com/pulumi/templates/pull/494, we applied the `SkipUpdate` setting to all templates in order ensure the test suite passed more reliably. (Up to that point, it had been failing for several months.) 

We should re-enable full-update tests for the most commonly used templates, starting with those used in the Getting Started guides. This includes TypeScript, JavaScript, Python, Go, C#, Java, and YAML for: 

- [x] AWS
- [x] Azure
- [x] GCP
- [x] Kubernetes

cnunciato:
Adds support for applying `integration.ProgramTestOptions` on a template-by-template basis, primarily to allow us to test a subset of templates (e.g., our core starter templates) with full updates and diff checks.

Note that the checks in this PR are failing (e.g., [here](https://github.com/pulumi/templates/actions/runs/4380318017/jobs/7667248950#step:21:644)) -- this is expected until https://github.com/pulumi/pulumi-aws/issues/2403 is fixed and any necessary template dependencies in this repo are updated.  

Fixes #533.

### DIFF
--- a/azure-python/__main__.py
+++ b/azure-python/__main__.py
@@ -5,21 +5,27 @@ from pulumi_azure_native import storage
 from pulumi_azure_native import resources
 
 # Create an Azure Resource Group
-resource_group = resources.ResourceGroup('resource_group')
+resource_group = resources.ResourceGroup("resource_group")
 
 # Create an Azure resource (Storage Account)
-account = storage.StorageAccount('sa',
+account = storage.StorageAccount(
+    "sa",
     resource_group_name=resource_group.name,
     sku=storage.SkuArgs(
         name=storage.SkuName.STANDARD_LRS,
     ),
-    kind=storage.Kind.STORAGE_V2)
+    kind=storage.Kind.STORAGE_V2,
+)
 
 # Export the primary key of the Storage Account
-primary_key = pulumi.Output.all(resource_group.name, account.name) \
-    .apply(lambda args: storage.list_storage_account_keys(
-        resource_group_name=args[0],
-        account_name=args[1]
-    )).apply(lambda accountKeys: accountKeys.keys[0].value)
+primary_key = (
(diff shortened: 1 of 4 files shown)


azure-python/__main__.py
```diff
@@ -5,21 +5,27 @@
 from pulumi_azure_native import resources
 
 # Create an Azure Resource Group
-resource_group = resources.ResourceGroup('resource_group')
```
cnunciato, line 8: These changes were necessary because the Windows test of this template was failing on a formatting issue.

tests/internal/testutils/templates.go
```diff
@@ -42,3 +43,68 @@ func FindAllTemplates(t *testing.T, templateUrl string) []TemplateInfo {
...
+		// Skip running a preview after the update, expecting no changes.
+		SkipPreview: skipFullTest,
+		// Skip running a stack export and import after the update.
+		SkipExportImport: skipFullTest,
+	}
+}
+
+// isGettingStartedTemplate returns whether the specified template is one that is used in
+// our Getting Started flows.
+func isGettingStartedTemplate(templateName string) bool {
```
justinvp, line 69: A potential alternative to maintaining the list in this function would be to base it on whether the template has `important: true`, which is the subset of templates shown by `pulumi new` by default, and also appears to currently align with the getting started list you have here. You can get at this information via `templateInfo.Template.Important`.

justinvp (approved):
One suggestion, but otherwise LGTM

tests/internal/testutils/templates.go
cnunciato: Oh, that's much nicer. I didn't realize that was there. Thanks for the tip!

PR #542, opened 2023-03-20
Use go 1.20x when running perf job

t0yv0:
Trying out the version of the job here: https://github.com/pulumi/templates/actions/runs/4472575517


t0yv0:
Previously the workflow was busted in the run step due to Go versions: https://github.com/pulumi/templates/actions/runs/4471372647/jobs/7856176823#step:19:55

t0yv0:
It looks like it's succeeding past the previous point of failure but good to wait out.

justinvp (approved):
Thanks!

PR #561, opened 2023-04-21
Update K8s version on AZ templates.

jkodroff:
Looks like 1.24 is no longer supported, at least in westus2.


scottslowe:
@jkodroff This is great, thank you! One small request: would you mind just removing the patch version entirely? Per the docs, it should then select the latest supported patch release for the specified `major.minor` version. That'll be one less thing we need to track/change regularly.

jkodroff:
I also `.gitignored` `.gradle` at the top level b/c VS Code was creating it automatically. (Did not add it at the template level because I don't know Gradle and it may be desirable to keep it under version control.)

jkodroff:
Per convo with @scottslowe, we're gonna omit the version entirely which will give us CURRENT_PROD_VERSION - 1, which will significantly reduce the maintenance burden for the template.

PR #570, opened 2023-05-02
Add .gradle to .gitignore

jkodroff:
Update container-azure-typescript to use the latest major version of
pulumi-docker and also ensure the template creates a linux/amd64 image
regardless of the host platform. (It appears Azure Container Apps only
support this platform.)


cnunciato (approved):
Sorry for the delay! Looks like these have since been updated -- but we'll definitely take that `.gradle` addition to the `.gitignore`. Thanks!

.gitignore
```diff
@@ -13,3 +13,4 @@ node_modules
 .vscode
 [Bb]in/
 [Oo]bj/
+.gradle
```
cnunciato, line 16:
```suggestion
.gradle

```

PR #587, opened 2023-05-25
Set azure native provider version to v2

### LINKED ISSUE #591: Update templates to use Azure Native v2
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details

As part of the v2 release, the templates that are using the Azure Native provider should be updated to use v2.

mnlumi:
resolves #591

### DIFF
--- a/azure-csharp/${PROJECT}.csproj
+++ b/azure-csharp/${PROJECT}.csproj
@@ -8,7 +8,7 @@
 
   <ItemGroup>
     <PackageReference Include="Pulumi" Version="3.*" />
-    <PackageReference Include="Pulumi.AzureNative" Version="2.0.0-alpha.1" />
+    <PackageReference Include="Pulumi.AzureNative" Version="2.0.0-alpha.4" />
   </ItemGroup>
 
 </Project>
--- a/azure-csharp/${PROJECT}.csproj
+++ b/azure-csharp/${PROJECT}.csproj
@@ -8,7 +8,7 @@
 
   <ItemGroup>
     <PackageReference Include="Pulumi" Version="3.*" />
-    <PackageReference Include="Pulumi.AzureNative" Version="1.*" />
+    <PackageReference Include="Pulumi.AzureNative" Version="2.0.0-alpha.1" />
   </ItemGroup>
 
 </Project>
--- a/azure-csharp/${PROJECT}.csproj
+++ b/azure-csharp/${PROJECT}.csproj
@@ -8,7 +8,7 @@
 
   <ItemGroup>
     <PackageReference Include="Pulumi" Version="3.*" />
-    <PackageReference Include="Pulumi.AzureNative" Version="2.0.0-beta.4" />
+    <PackageReference Include="Pulumi.AzureNative" Version="2.0.0" />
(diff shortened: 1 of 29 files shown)


azure-csharp/${PROJECT}.csproj
```diff
@@ -8,7 +8,7 @@
 
   <ItemGroup>
     <PackageReference Include="Pulumi" Version="3.*" />
-    <PackageReference Include="Pulumi.AzureNative" Version="1.*" />
+    <PackageReference Include="Pulumi.AzureNative" Version="2.0.0" />
```
iwahbe, line 11: Why is this not `"2.*"`, like `"1.*"` before?

thomas11: good catch!

azure-javascript/package.json
```diff
@@ -3,6 +3,6 @@
     "main": "index.js",
     "dependencies": {
         "@pulumi/pulumi": "^3.0.0",
-        "@pulumi/azure-native": "^1.0.0"
+        "@pulumi/azure-native": "v2.0.0"
```
iwahbe, line 6: Why have we decided to pin `v2.0.0` instead of any value in `v2.*` (which is what `^1.0.0` did before)?

This is the same question as for the .NET version.

azure-python/requirements.txt
```diff
@@ -1,2 +1,2 @@
 pulumi>=3.0.0,<4.0.0
-pulumi-azure-native>=1.0.0,<2.0.0
+pulumi-azure-native==v2.0.0
```
iwahbe, line 2: Again, why not:
```
pulumi-azure-native>=v2.0.0,<3.0.0
```

container-azure-go/go.mod
```diff
@@ -3,12 +3,12 @@ module ${PROJECT}
...
-	github.com/pulumi/pulumi-azure-native-sdk/containerinstance v1.88.1
-	github.com/pulumi/pulumi-azure-native-sdk/containerregistry v1.88.1
-	github.com/pulumi/pulumi-azure-native-sdk/resources v1.88.1
+	github.com/pulumi/pulumi-azure-native-sdk/containerinstance/v2 v2.0.0
+	github.com/pulumi/pulumi-azure-native-sdk/containerregistry/v2 v2.0.0
+	github.com/pulumi/pulumi-azure-native-sdk/resources/v2 v2.0.0
 	github.com/pulumi/pulumi-random/sdk/v4 v4.8.2
 	github.com/pulumi/pulumi-docker/sdk/v4 v4.2.3
-	github.com/pulumi/pulumi/sdk/v3 v3.42.0
+	github.com/pulumi/pulumi/sdk/v3 v3.72.0
```
iwahbe, line 11: This bumps to `v3.72.0` but `azure-go/go.mod` bumps to `v3.74.0`. Any reason for the inconsistency?

iwahbe (approved):
I left a couple of version nit-picks. The rest looks right.

Issue #588, opened 2023-05-26
Update templates to use docker v4


mnlumi:
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details

All templates using pulumi-docker should be updated to v4 to use the latest features and functionality.

PR #589, opened 2023-05-26
Update pulumi-docker version to v4.2.3

mnlumi:
Update pulumi-docker version to v4.2.3

Resolves https://github.com/pulumi/templates/issues/588


mnlumi:
Abandoning in favor of https://github.com/pulumi/templates/pull/593

Issue #567, opened 2023-05-01
Update all templates to use Docker provider 4.x


jkodroff:
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details

At the very least, the Azure Container template is still using version 3.x of the Docker provider. There have been major improvements in version 4.x of the provider, so we should update to the latest major version.

scottslowe:
@mnlumi I don't think `kubernetes-aws-go` is affected. The `container-aws-*` templates, though, do use AWSX.

mnlumi:
@scottslowe I do see the template pulls in docker v3 as an indirect dependency when it runs. This template will need to be updated when awsx is updated in the future. I can put kubernetes-aws-go on a separate issue and leave this one for direct docker v3 updates.

scottslowe:
@mnlumi The `container-azure-*` and `container-gcp-*` templates have been updated (#593 was merged). Do we want to keep this issue open for the `container-aws-*` templates, which use Docker 3.x via AWSX?

lukehoban:
> Do we want to keep this issue open for the container-aws-* templates, which use Docker 3.x via AWSX?

I think we can close this, and track the rest as part of updating awsx to docker v4.

PR #593, opened 2023-06-08
Update templates to docker v4

mnlumi:
Addresses https://github.com/pulumi/templates/issues/567

Update templates to use docker v4 and set standard platform.

### DIFF
--- a/container-azure-csharp/Program.cs
+++ b/container-azure-csharp/Program.cs
@@ -44,6 +44,7 @@ return await Pulumi.Deployment.RunAsync(() =>
         ImageName = Pulumi.Output.Format($"{registry.LoginServer}/{imageName}:{imageTag}"),
         Build = new Docker.Inputs.DockerBuildArgs {
             Context = appPath,
+            Platform = "linux/amd64",
         },
         Registry = new Docker.Inputs.RegistryArgs {
             Server = registry.LoginServer,
--- a/container-azure-csharp/Program.cs
+++ b/container-azure-csharp/Program.cs
@@ -42,10 +42,10 @@ return await Pulumi.Deployment.RunAsync(() =>
     var image = new Docker.Image("image", new()
     {
         ImageName = Pulumi.Output.Format($"{registry.LoginServer}/{imageName}:{imageTag}"),
-        Build = new Docker.DockerBuild {
+        Build = new Docker.Inputs.DockerBuildArgs {
             Context = appPath,
         },
-        Registry = new Docker.ImageRegistry {
+        Registry = new Docker.Inputs.RegistryArgs {
             Server = registry.LoginServer,
             Username = registryUsername,
             Password = registryPassword,
--- a/container-azure-go/main.go
+++ b/container-azure-go/main.go
@@ -145,6 +145,9 @@ func main() {
 				},
 			},
 		})
+		if err != nil {
+			return err
+		}
 
 		// Export the service's IP address, hostname, and fully-qualified URL.
 		ctx.Export("ip", containerGroup.IpAddress.Elem().Ip())
--- a/container-azure-go/main.go
+++ b/container-azure-go/main.go
@@ -74,7 +74,7 @@ func main() {
 			ImageName: pulumi.Sprintf("%s/%s:%s", registry.LoginServer, imageName, imageTag),
 			Build: docker.DockerBuildArgs{
 				Context: pulumi.String(appPath),
-				Platform: pulumi.String("linux/arm64"),
+				Platform: pulumi.String("linux/amd64"),
 			},
 			Registry: docker.ImageRegistryArgs{
 				Server:   registry.LoginServer,
--- a/container-azure-go/main.go
+++ b/container-azure-go/main.go
@@ -7,7 +7,7 @@ import (
 	"github.com/pulumi/pulumi-azure-native-sdk/containerinstance"
 	"github.com/pulumi/pulumi-azure-native-sdk/containerregistry"
 	"github.com/pulumi/pulumi-azure-native-sdk/resources"
-	"github.com/pulumi/pulumi-docker/sdk/v3/go/docker"
+	"github.com/pulumi/pulumi-docker/sdk/v4/go/docker"
 	"github.com/pulumi/pulumi-random/sdk/v4/go/random"
 	"github.com/pulumi/pulumi/sdk/v3/go/pulumi"
 	"github.com/pulumi/pulumi/sdk/v3/go/pulumi/config"
@@ -74,6 +74,7 @@ func main() {
 			ImageName: pulumi.Sprintf("%s/%s:%s", registry.LoginServer, imageName, imageTag),
 			Build: docker.DockerBuildArgs{
 				Context: pulumi.String(appPath),
+				Platform: pulumi.String("linux/arm64"),
 			},
 			Registry: docker.ImageRegistryArgs{
 				Server:   registry.LoginServer,
--- a/container-azure-go/main.go
+++ b/container-azure-go/main.go
@@ -76,7 +76,7 @@ func main() {
 				Context: pulumi.String(appPath),
 				Platform: pulumi.String("linux/amd64"),
 			},
-			Registry: docker.ImageRegistryArgs{
+			Registry: docker.RegistryArgs{
 				Server:   registry.LoginServer,
 				Username: registryUsername,
 				Password: registryPassword,
--- a/container-gcp-csharp/Program.cs
+++ b/container-gcp-csharp/Program.cs
@@ -1,10 +1,9 @@
-﻿// Before running `pulumi up`, configure Docker for Artifact Registry authentication
-// as described here: https://cloud.google.com/artifact-registry/docs/docker/authentication
-
 using Pulumi;
 using Gcp = Pulumi.Gcp;
 using Docker = Pulumi.Docker;
 using System.Collections.Generic;
+using Random = Pulumi.Random;
+using Output = Pulumi.Output;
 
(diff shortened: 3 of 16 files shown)


scottslowe (changes requested):
The suggestion I supplied for `container-azure-go` probably needs to be applied to `container-gcp-go` as well. Additionally, we should add a block to check the value of `err` when creating the container service in `container-azure-go`, since that value of `err` is never checked. I don't know if the GCP version needs the same or not. Python and TypeScript seem to work great; I didn't test the C# templates.

container-azure-go/main.go
```diff
@@ -74,6 +74,7 @@ func main() {
 			ImageName: pulumi.Sprintf("%s/%s:%s", registry.LoginServer, imageName, imageTag),
 			Build: docker.DockerBuildArgs{
 				Context: pulumi.String(appPath),
+				Platform: pulumi.String("linux/amd64"),
 			},
 			Registry: docker.ImageRegistryArgs{
```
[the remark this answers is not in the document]
mnlumi: Good catch @scottslowe! I had that in my test program and missed copying it over. Thanks!

As for the `err` check, I'm not familiar enough with go for that. Happy to pair or a receive a commit to my branch. 😄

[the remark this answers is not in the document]
mnlumi: @scottslowe I added the err check. Please re-review.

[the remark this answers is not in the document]
scottslowe: @mnlumi I was going to reach out to you yesterday and see if you wanted to pair really quick on the err check; sorry I didn't get to it! I'll re-review shortly.

scottslowe (approved):
Approved; thanks! Note that I did not personally test the C# templates, but I did test the TypeScript, Python, and Golang ones. Given the importance of the templates in the user experience flow with `pulumi new`, I highly recommend that someone directly tests the C# templates to make sure they work as expected before merging (as a backup to the existing testing infrastructure).

container-azure-csharp/Program.cs
```diff
@@ -42,10 +42,11 @@
     var image = new Docker.Image("image", new()
     {
         ImageName = Pulumi.Output.Format($"{registry.LoginServer}/{imageName}:{imageTag}"),
-        Build = new Docker.DockerBuild {
+        Build = new Docker.Inputs.DockerBuildArgs {
```
cnunciato, line 45: When I run this, I get:

```
MSBuild version 17.5.1+f6fdcf537 for .NET
  Determining projects to restore...
  Restored /Users/cnunciato/Projects/dev/monica-csharp-pr/monica-csharp-pr.csproj (in 1.27 sec).
/Users/cnunciato/Projects/dev/monica-csharp-pr/Program.cs(45,35): error CS0234: The type or namespace name 'DockerBuildArgs' does not exist in the namespace 'Pulumi.Docker.Inputs' (are you missing an assembly reference?) [/Users/cnunciato/Projects/dev/monica-csharp-pr/monica-csharp-pr.csproj]
/Users/cnunciato/Projects/dev/monica-csharp-pr/Program.cs(49,38): error CS0234: The type or namespace name 'RegistryArgs' does not exist in the namespace 'Pulumi.Docker.Inputs' (are you missing an assembly reference?) [/Users/cnunciato/Projects/dev/monica-csharp-pr/monica-csharp-pr.csproj]

Build FAILED.
```

I think this is because the `.csproj` file needs to be updated -- will see if that's the case.

cnunciato: Yeah that seems to be it -- in `{PROJECT}.csproj`, just change that `3.*` to `4.*`.

container-gcp-typescript/index.ts
```diff
@@ -21,6 +21,7 @@ const image = new docker.Image("image", {
     imageName: `gcr.io/${project}/${imageName}`,
     build: {
         context: appPath,
+        platform: "linux/amd64",
```
cnunciato, line 24: I get a compile error on this one: 

```
error: Running program '/Users/cnunciato/Projects/dev/monica-csharp-pr/' failed with an unhandled exception:
    TSError: ⨯ Unable to compile TypeScript:
    index.ts(25,9): error TS2322: Type '{ context: string; platform: string; env: { DOCKER_DEFAULT_PLATFORM: string; }; }' is not assignable to type 'DockerBuild | Promise<DockerBuild> | OutputInstance<DockerBuild> | undefined'.
      Object literal may only specify known properties, and 'env' does not exist in type 'Input<DockerBuild>'
```

Will see if I can figure it out.

cnunciato: Ah, presumably because `env` (which follows) is no longer on the API.

cnunciato (changes requested):
Couple things to fix up before we merge, but on the whole looks good! (I did test each template here, and the issues I noticed are noted above.) The GCP templates do seem to fail with this error:

```
docker:index:Image (image):
  error: unauthorized: You don't have the needed permissions to perform this operation, and you may have invalid credentials. To authenticate your request, follow the steps in: https://cloud.google.com/container-registry/docs/advanced-authentication
```

... but I get this with the current set of GCP templates also, so this seems to be unrelated to the content of this PR.

scottslowe:
I spent some time tonight trying to resolve the issue with the GCP templates (per @cnunciato's comment above). The error seen by @cnunciato is apparently related to a misconfiguration (or lack of configuration) in Docker. Docker needs to be configured to use `docker-credential-gcloud` in `config.json`, or else the error shown above is returned.

Unfortunately, it appears that even with the configuration an error still occurs:
```
  gcp:cloudrun:Service (service):
    error: 1 error occurred:
    	* Error creating Service: Post "https://us-central1-run.googleapis.com/apis/serving.knative.dev/v1/namespaces/pulumi-devrel/services?alt=json": oauth2: cannot fetch token: 400 Bad Request
    Response: {
      "error": "invalid_grant",
      "error_description": "Token has been expired or revoked."
    }```
Running `gcloud auth revoke` followed by `gcloud auth login` does not resolve the errors. The Cloud Run API is enabled. I'm not really sure what's happening here.

**Comment by scottslowe** [comment]
It would appear that we need to switch the templates to use Artifact Registry instead of Container Registry; however, I've run into an issue using `gcp.artifactregistry.Repository`. This error occurs:
```
  gcp:artifactregistry:Repository (my-repo):
    error: 1 error occurred:
    	* Error creating Repository: Post "https://artifactregistry.googleapis.com/v1/projects/pulumi-devrel/locations/us-central1/repositories?alt=json&repository_id=my-repo": oauth2: "invalid_grant" "Token has been expired or revoked."
```

File: container-gcp-csharp/Program.cs [hunk 4]
```diff
@@ -19,12 +22,27 @@
...
+    var repository = new Gcp.ArtifactRegistry.Repository("my-repo", new()
+    {
+        Description = "Repository for container image",
+        Format = "DOCKER",
+        Location = location,
+        RepositoryId = "my-repo",
+    });
+
+    // Form the repository URL
+    var repoUrl = $"{location}-docker.pkg.dev/{project}/my-repo";
```
cnunciato, line 35: It's probably worth pointing out that this failed the first time I ran it because `us-central1-docker.pkg.dev/my-teams-shared-project/my-repo` had already been created by someone else:

```
gcp:artifactregistry:Repository (my-repo):
    error: 1 error occurred:
        * Error creating Repository: googleapi: Error 409: the repository already exists
```

I'm not sure how folks use GCP in practice, but if projects are commonly shared among multiple developers, I suspect this'll fail pretty regularly -- probably even for us, when we start running templates in automated tests.

Since the assumption seems to be that the repo will be managed by the Pulumi program, should we consider naming the repo somehow randomly to avoid collisions? What do you think?

cnunciato (approved):
A few suggestions to consider, but nothing we couldn't come back to in a follow-up. Nice work on this one, by the way! I know it was quite the multi-layered onion. 🧅

container-gcp-csharp/Program.cs
scottslowe: You're right. Unfortunately, we can't use Pulumi's auto-naming functionality here. However, I've implemented a random string approach (in commits 9628dd5 and 75ba1c3) that should help. PTAL!

container-gcp-csharp/Program.cs
```diff
@@ -1,4 +1,7 @@
-﻿using Pulumi;
+﻿// Before running `pulumi up`, configure Docker for Artifact Registry authentication
+// as described here: https://cloud.google.com/artifact-registry/docs/docker/authentication
```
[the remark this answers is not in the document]
scottslowe: Resolved in latest commits

container-gcp-go/main.go
```diff
@@ -4,11 +4,14 @@ import (
...
 
 	"github.com/pulumi/pulumi-docker/sdk/v4/go/docker"
+	"github.com/pulumi/pulumi-gcp/sdk/v6/go/gcp/artifactregistry"
 	"github.com/pulumi/pulumi-gcp/sdk/v6/go/gcp/cloudrun"
 	"github.com/pulumi/pulumi/sdk/v3/go/pulumi"
 	"github.com/pulumi/pulumi/sdk/v3/go/pulumi/config"
 )
 
+// Before running `pulumi up`, configure Docker for authentication to Artifact Registry as
+// described here: https://cloud.google.com/artifact-registry/docs/docker/authentication
```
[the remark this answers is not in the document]
scottslowe: Resolved in latest commits

cnunciato (approved):
Looks good. Thanks for the revisions! 🚀

mnlumi:
Thank you @scottslowe and @cnunciato for getting this across the finish line! This ended up being much more involved than anticipated.

PR #598, opened 2023-06-14
Upgrades all Pulumi AWS requirements to 6.0.0-alpha.1

kpitzen:
Note - AWSX and API Gateway are not yet updated to depend on v6


lukehoban:
This is now tracked in #619.

PR #608, opened 2023-07-10
[DO NOT MERGE] Update to Azure Native v2

mikhailshilkov:


mikhailshilkov:
Oh, I'll update #587 instead

PR #612, opened 2023-07-14
Adjust azure-native version deps per feedback

mnlumi:
From @iwahbe feedback here: https://github.com/pulumi/templates/pull/587#pullrequestreview-1530246962

### DIFF
--- a/azure-javascript/package.json
+++ b/azure-javascript/package.json
@@ -3,6 +3,6 @@
     "main": "index.js",
     "dependencies": {
         "@pulumi/pulumi": "^3.0.0",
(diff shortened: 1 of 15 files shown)


thomas11 (changes requested):
Thank you for fixing up what I missed!

Searching for "2.0.0" through the repo, I found two more:

```
vm-azure-python/requirements.txt
2:pulumi-azure-native>=1.0.0,<2.0.0

static-website-azure-python/requirements.txt
2:pulumi-azure-native>=1.0.0,<2.0.0
```

thomas11 (approved):

PR #613, opened 2023-07-15
Update templates for kubernetes v4

mnlumi:


mnlumi:
@scottslowe or @cnunciato - Can we get the kubernetes tests to run instead of being skipped?

mnlumi:
Thanks so much @cnunciato!

kubernetes-typescript/package.json
```diff
@@ -6,7 +6,7 @@
     },
     "dependencies": {
         "@pulumi/pulumi": "^3.0.0",
-        "@pulumi/kubernetes": "^3.0.0",
+        "@pulumi/kubernetes": "^4.0.0",
         "@pulumi/kubernetesx": "^0.1.5"
```
stas-sbi, line 10: After this change kubernetes-typescript template doesn't work because kubernetesx conflicts with kubernetes 4.x https://github.com/pulumi/pulumi-kubernetesx/blob/097ea75f38fc7167894382e431bd0cf5ac7a614d/nodejs/kubernetesx/package.json#L24

kubernetes-typescript/package.json
```diff
@@ -6,7 +6,7 @@
     },
     "dependencies": {
         "@pulumi/pulumi": "^3.0.0",
-        "@pulumi/kubernetes": "^3.0.0",
```
stas-sbi, line 9: After this change kubernetes-typescript template doesn't work because kubernetesx conflicts with kubernetes 4.x https://github.com/pulumi/pulumi-kubernetesx/blob/097ea75f38fc7167894382e431bd0cf5ac7a614d/nodejs/kubernetesx/package.json#L24

giner, line 9: After this change kubernetes-typescript template doesn't work because kubernetesx conflicts with kubernetes 4.x https://github.com/pulumi/pulumi-kubernetesx/blob/097ea75f38fc7167894382e431bd0cf5ac7a614d/nodejs/kubernetesx/package.json#L24

cnunciato: @giner I just tested this template (Node 18.15.0, Pulumi 3.76.1) and it works for me. Can you share more what issues you're seeing?

cnunciato: Looks like template install work with `yarn`, but fails with `npm`. We're looking into it.

cnunciato: Should be fixed with https://github.com/pulumi/templates/pull/626.

PR #614, opened 2023-07-18
chore: bump to node 16

simenandre:
There was a few projects that were not bumped to node 16. Not sure why, so though I'd bump 'em.


cnunciato (approved):
🙌

PR #615, opened 2023-07-18
feat: use engines for all node templates

simenandre:
The benefit of using `"engines"` is that developers are told which version of Node they should use.

Even though the `pulumi/pulumi` package is using `>=8.13.0 || >=10.10.0`, I think pushing developers to at least Node v16 makes sense.

Related: We should consider using Node LTS for all the templates. It's a good practice to use active LTS on new Node projects.


simenandre:
Lets close this!

PR #616, opened 2023-07-18
chore: bump node to active LTS

simenandre:
Node recommends using Long Term Support (LTS) version for most users. We've kept the template updated to one version behind for quite some time. IMHO, I'm not sure we should be one version behind LTS.

Node has _current_ which is v20 at the time of writing. That feels too unstable for Pulumi templates, but LTS is what is mostly recommended. LTS is kind of _one version behind_ 😊

See the schedule here:
https://github.com/nodejs/release#release-schedule

closes #614

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -38,7 +38,7 @@ jobs:
       matrix:
         platform: [ubuntu-latest, macOS-latest, windows-latest]
         go-version: [1.18.x]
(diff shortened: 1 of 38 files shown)


cnunciato:
Hi Simen! Big apologies for the delay in getting to this. We've since upgraded to Node 18 across the board, so I think we should close this PR for now. Thanks as always for your contributions .. and patience! 😅

PR #619, opened 2023-07-25
Update templates to AWS 6.0

mnlumi:
Resolves https://github.com/pulumi/templates/issues/618

### DIFF
--- a/aws-go/go.mod
+++ b/aws-go/go.mod
@@ -3,6 +3,6 @@ module ${PROJECT}
 go 1.18
 
 require (
-	github.com/pulumi/pulumi-aws/sdk/v5 v5.33.0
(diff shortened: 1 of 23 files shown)


cnunciato:
@mnlumi Do we want to update these to an alpha release? I would think probably not, so seemed worth asking.

mnlumi:
@cnunciato These are intentionally set to the alpha release for testing purposes. It has the `do-not-merge` label on until we are ready to release 6.0 and will be updated to the full release prior to merging.

cnunciato:
@mnlumi Excellent, thanks for confirming!

mnlumi:
> @mnlumi Excellent, thanks for confirming! Feel free to switch it to Draft as well in the meantime.

The tests don't run when the PR is in draft. I've remove draft status, but still have the 'do-not-merge' label on.

PR #624, opened 2023-08-02
Prevent java template from using 6.0 alpha versions

mnlumi:


mnlumi:
Related to https://github.com/pulumi/pulumi-aws/issues/2658

mnlumi:
The latest version of aws is 5.4. I've validated that this works as expected.

```
Do you want to perform this update? details
+ pulumi:pulumi:Stack: (create)
    [urn=urn:pulumi:dev::aws-java-2::pulumi:pulumi:Stack::aws-java-2-dev]
    + aws:s3/bucket:Bucket: (create)
        [urn=urn:pulumi:dev::aws-java-2::aws:s3/bucket:Bucket::my-bucket]
        [provider=urn:pulumi:dev::aws-java-2::pulumi:providers:aws::default_5_4_0::......]
        acl         : "private"
        bucket      : "my-bucket-f3a10fd"
        forceDestroy: false
    --outputs:--
    bucketName: "my-bucket-f3a10fd"
```

mnlumi:
Alpha versions are considered less than the numeric version so even with `<version>[5.0.0,6.0.0]</version>`, the 6.0.0-alpha will be selected.

t0yv0:
This look's good 🙏 

It doesn't shield users from versions such as https://repo1.maven.org/maven2/com/pulumi/aws/5.43.0-alpha.1689671883+f787a47a/ but those are not currently known to be affected.

mnlumi:
> This look's good 🙏
> 
> It doesn't shield users from versions such as https://repo1.maven.org/maven2/com/pulumi/aws/5.43.0-alpha.1689671883+f787a47a/ but those are not currently known to be affected.

True, I could set it to 5.4 explicitly, but didn't want to eliminate 5.5 from being used in case there was some other emergent release. My goal is to prevent the next major version's alpha releases from being consumed.

mnlumi:
@cnunciato I'm not sure why this test is failing. Could you help take a look please? 
```
  error: error: no changes were expected but changes were proposed
```
Also, note the scaleway update due to failing tests around go being 1.18 max expected but 1.20 in the actual program.

cnunciato:
@mnlumi It looks like the `aws-java` test [passed](https://github.com/pulumi/templates/actions/runs/5743247020/job/15567115572#step:21:209) for your [first commit](https://github.com/pulumi/templates/pull/624/commits/0e6ec11479eab3b168d98dd41f3de3fc2e754266), but [started failing](https://github.com/pulumi/templates/actions/runs/5743583708/job/15568153639#step:21:209) with the second one and still is. From the output, it looks like follow-up updates to the templated program [produce a diff](https://github.com/pulumi/templates/actions/runs/5743583708/job/15568153639#step:21:249), which they shouldn't:

image

From the checkpoint it looks like the version that failed (i.e., that produced the diff) was 5.4.0.

aws-java/pom.xml
```diff
@@ -26,7 +26,7 @@
         <dependency>
             <groupId>com.pulumi</groupId>
             <artifactId>aws</artifactId>
-            <version>(,6.0.0]</version>
+            <version>(,5.9]</version>
```
ringods, line 29: I would use `5.99`. Current AWS provider release is already at 5.42.0.

ringods (changes requested):

aws-java/pom.xml
mnlumi: Thanks @ringods for catching this!

cnunciato (approved):

PR #644, opened 2023-09-11
Unset the important flag on the SageMaker template

cnunciato:
We currently only use this flag for our tier-1 templates (e.g., `aws-typescript` and the like). Removing it from this template will also "fix" the [failing tests](https://github.com/pulumi/templates/actions/runs/6127351193/job/16632983679) (i.e., by skipping them).


scottslowe (approved):
Approved

PR #645, opened 2023-09-12
Fix kubernetes-go template

cnunciato:
The template uses a method (`deployment.Metadata.Elem()`) which is apparently no longer valid.

Part of https://github.com/pulumi/pulumi-hugo/issues/3354.


scottslowe (approved):
Approved, test failures are all related to `sagemaker-aws-python`

PR #650, opened 2023-09-19
Upgrade templates to aws v6, awsx v2, and eks v2

thomas11:
Upgrade templates to aws v6, awsx v2, and eks v2

### DIFF
--- a/aws-javascript/package.json
+++ b/aws-javascript/package.json
@@ -4,6 +4,6 @@
     "dependencies": {
         "@pulumi/pulumi": "^3.0.0",
         "@pulumi/aws": "^6.0.0",
-        "@pulumi/awsx": "2.0.0-beta.3"
+        "@pulumi/awsx": "^2.0.2"
     }
 }
--- a/aws-javascript/package.json
+++ b/aws-javascript/package.json
@@ -3,7 +3,7 @@
     "main": "index.js",
     "dependencies": {
         "@pulumi/pulumi": "^3.0.0",
-        "@pulumi/aws": "^5.0.0",
-        "@pulumi/awsx": "^1.0.0"
+        "@pulumi/aws": "^6.0.0",
+        "@pulumi/awsx": "2.0.0-beta.3"
     }
 }
--- a/container-aws-csharp/${PROJECT}.csproj
+++ b/container-aws-csharp/${PROJECT}.csproj
@@ -8,8 +8,8 @@
 
   <ItemGroup>
     <PackageReference Include="Pulumi" Version="3.*" />
-    <PackageReference Include="Pulumi.Aws" Version="5.*" />
-    <PackageReference Include="Pulumi.Awsx" Version="1.*" />
+    <PackageReference Include="Pulumi.Aws" Version="6.*" />
+    <PackageReference Include="Pulumi.Awsx" Version="2.0.0-beta.3" />
   </ItemGroup>
 
 </Project>
--- a/container-aws-csharp/${PROJECT}.csproj
+++ b/container-aws-csharp/${PROJECT}.csproj
@@ -9,7 +9,7 @@
   <ItemGroup>
     <PackageReference Include="Pulumi" Version="3.*" />
     <PackageReference Include="Pulumi.Aws" Version="6.*" />
-    <PackageReference Include="Pulumi.Awsx" Version="2.0.0-beta.3" />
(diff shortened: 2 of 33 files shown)


cnunciato (commented):
I don't think we want to use alpha/beta versions of these now that they're all post-1.0. Is there some reason we should?

aws-javascript/package.json
```diff
@@ -3,7 +3,7 @@
     "main": "index.js",
     "dependencies": {
         "@pulumi/pulumi": "^3.0.0",
-        "@pulumi/aws": "^5.0.0",
-        "@pulumi/awsx": "^1.0.0"
+        "@pulumi/aws": "^6.0.0",
+        "@pulumi/awsx": "2.0.0-beta.3"
```
iwahbe, line 7: Is it possible to run acceptance tests on the PR and then merge only after awsx has come out?

iwahbe (commented):
I'm also nervous about baking in beta versions into templates.

aws-javascript/package.json
thomas11: Yes, that's the plan.

thomas11:
> I don't think we want to use alpha/beta versions of these now that they're all post-1.0. Is there some reason we should?

We won't, they are simply all I have right now before awsx/eks are released.

container-aws-csharp/${PROJECT}.csproj
```diff
@@ -8,8 +8,8 @@
 
   <ItemGroup>
     <PackageReference Include="Pulumi" Version="3.*" />
-    <PackageReference Include="Pulumi.Aws" Version="5.*" />
-    <PackageReference Include="Pulumi.Awsx" Version="1.*" />
+    <PackageReference Include="Pulumi.Aws" Version="6.*" />
+    <PackageReference Include="Pulumi.Awsx" Version="2.0.0-beta.3" />
```
t0yv0, line 12: Ah this is hard. We're changing this away from beta before merging yes? I feel like this is so easy to let slip through.

container-aws-python/requirements.txt
```diff
@@ -1,3 +1,3 @@
 pulumi>=3.0.0,<4.0.0
-pulumi-aws>=5.10.0,<6.0.0
-pulumi-awsx>=1.0.0,<2.0.0
+pulumi-aws>=6.0.0,<7.0.0
+pulumi-awsx==2.0.0b3
```
t0yv0, line 3: Another beta.

kubernetes-aws-csharp/${PROJECT}.csproj
```diff
@@ -8,9 +8,9 @@
 
   <ItemGroup>
     <PackageReference Include="Pulumi" Version="3.*" />
-    <PackageReference Include="Pulumi.Aws" Version="5.*" />
-    <PackageReference Include="Pulumi.Awsx" Version="1.*" />
-    <PackageReference Include="Pulumi.Eks" Version="1.*" />
+    <PackageReference Include="Pulumi.Aws" Version="6.*" />
+    <PackageReference Include="Pulumi.Awsx" Version="2.0.0-beta.3" />
```
t0yv0, line 12: Ditto

serverless-aws-csharp/${PROJECT}.csproj
```diff
@@ -8,8 +8,8 @@
 
 	<ItemGroup>
 		<PackageReference Include="Pulumi" Version="3.*" />
-		<PackageReference Include="Pulumi.Aws" Version="5.*" />
-		<PackageReference Include="Pulumi.AwsApiGateway" Version="1.*" />
+		<PackageReference Include="Pulumi.Aws" Version="6.*" />
+		<PackageReference Include="Pulumi.AwsApiGateway" Version="1.1.0-alpha.3" />
```
scottslowe, line 12: Why is a different version specified here for C#, when the other language versions for this architecture specify version 2.x?

scottslowe (changes requested):

serverless-aws-csharp/${PROJECT}.csproj
thomas11: An oversight, thank you for spotting it!

scottslowe (approved):
Approved, thank you!

iwahbe (approved):

PR #666, opened 2023-10-17
Update generate-go-mod.sh to compute go.mod/go.sum

blampe:
This updates the script responsible for bumping packages in `go.mod` to also run `go mod tidy` along the way. This ensures that the projects we template are usable out-of-the-box without requiring the user to run `go mod tidy` themselves.

`go mod tidy` can't parse `go.mod` files of the form `module ${PACKAGE}`, so we temporarily re-write `${PACKAGE}`, perform the tidy, and then restore `${PACKAGE}`.

Refs https://github.com/pulumi/pulumi-cloud-requests/issues/312 
Refs https://github.com/pulumi/templates/issues/114

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -1,0 +1,1 @@
+export GOTOOLCHAIN=go1.20.10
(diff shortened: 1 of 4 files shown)


cnunciato:
Ok to close this one also?

PR #667, opened 2023-10-17
Include concrete go.mod/go.sum files

blampe:
This generates concrete `go.mod` and `go.sum` files so templated projects are usable without requiring the user to run `go mod tidy`.

Refs https://github.com/pulumi/templates/pull/666
Refs https://github.com/pulumi/pulumi-cloud-requests/issues/312 
Refs https://github.com/pulumi/templates/issues/114


cnunciato (commented):
I'm generally disinclined to add these here -- for one, because we ignore lockfiles for other languages (presumably to allow consumers to pick up transitive deps without our involvement), but also because `pulumi new` produces these files automatically. How would adding them here be "better"?

blampe:
@cnunciato 

> for one, because we ignore lockfiles for other languages

Other languages are happy to re-construct a lock file from scratch, but Go is more strict. A JS project is still runnable with only a `package.json`, but a Go project won't compile unless it has a `go.sum`.

I think it's reasonable to expect that after templating a new project (that is, after performing all the necessary string replacements), the final output should be a runnable program.

We _do_ include `go.sum` in https://github.com/pulumi/examples, so it's not clear to me why templates should be different.

> but also because `pulumi new` produces these files automatically

`pulumi new` performs a `go mod tidy` as a workaround to the incomplete Go programs we're templating. It doesn't _need_ to do that if the template produces a valid program instead. 

Importantly, relying on `tidy` like this requires a Go language runtime to be present, which we don't have in the NPW (https://github.com/pulumi/pulumi-cloud-requests/issues/312). AFAICT Go is the only language that needs a runtime like this, and this PR would remove that requirement.

> (presumably to allow consumers to pick up transitive deps without our involvement)

Delaying dependency resolution until template-time leaves the door open for frustrating version incompatibilities. Why risk exposing that to the user?

We already know which transitive dependencies you'll need -- we resolve them in order to test the templates, but then we discard that information. By including it in `go.sum` instead, no matter when you perform your templating, as long as those dependencies are still available your project will build. 

> How would adding them here be "better"?

* Removes the language runtime requirement for Go.
* Consistency with pulumi/examples.
* Makes dependencies explicit and auditable.

cnunciato:
Been chatting with Bryce out of band on this. The motivation for this PR is two-part: first, to make Deployments of templated Go projects work (they currently don't, because Deployments doesn't run `go mod tidy` to resolve dependencies, which `pulumi new` does implicitly -- I've submitted a Pulumi Cloud PR to address this); and second, to make successive deployments more reliable by checking in lockfiles. We're currently discussing adding lockfiles to the templates repo in general ([thread here](https://pulumi.slack.com/archives/CBUP505U7/p1698715058158329)), so this PR is essentially waiting on the outcome of that discussion (as we'd probably want to handle all languages similarly).

cnunciato:
@blampe Ok to close?

cnunciato:
Closing this just to tidy up a bit, but feel free to reopen if you need.

PR #680, opened 2023-11-08
Update pulumi-gcp to v7.0.0

guineveresaenger:
Update all dependencies to use GCP v7.0.0 🎉 

Fixes https://github.com/pulumi/pulumi-gcp/issues/1250.


iwahbe (approved):
I'm a bit worried about the test failures, but I'm seeing the same failures on AWS, and only on windows. I would confirm that this isn't the first time we have seen the windows tests fail here. If the failure is common, then I'm ok merging with failed tests.

guineveresaenger:
I have not seen these failures so I'm rerunning that suite.

PR #683, opened 2023-11-15
Add SSH key to enable testing

scottslowe:
This PR adds an SSH key to the "VM on Azure" architecture templates. Adding the SSH key enables automated testing of the templates, which are currently excluded from automated testing.

The PR includes the following changes:
* Updates templates to add an SSH key resource
* Uses the SSH key resource for the VM resource
* Removes the SSH key as a required configuration value


desteves (approved):
I tested the `vm-azure-go`

```bash

mkdir testrun 
pulumi new https://github.com/pulumi/templates/tree/feat/slowe/add-ssh-key-vm-azure/vm-azure-go  --dir testrun
cd testrun 
az login
pulumi up -y
pulumi stack output privatekey --show-secrets
# shows the privatekey value
pulumi destroy -y
```

image

PR #686, opened 2023-11-16
Update to dotnet 8

Frassle:
.NET 8 is an LTS release. We should update templates to use this rather than 6.

N.B. We didn't update templates for 7 because it wasn't an LTS release.


tgummerer (approved):

scottslowe (approved):
Changes look correct and tests are passing, so approved

PR #693, opened 2023-11-28
Fix policy version string

cnunciato:


scottslowe (approved):
Approved, assuming the failing macOS test isn't related

PR #695, opened 2023-11-30
Update GCP cloudrun template to use docker image repo digest identifier

mnlumi:
Update GCP cloudrun template to use docker image repo digest identifier

By using image name in the cloud run service, the image won't be redeployed when the app changes currently.

Ref:
https://github.com/pulumi/pulumi-gcp/issues/1166
https://github.com/pulumi/pulumi-docker/issues/726
https://github.com/pulumi/pulumi-docker/pull/739

### DIFF
--- a/container-gcp-go/main.go
+++ b/container-gcp-go/main.go
@@ -97,7 +97,7 @@ func main() {
 				Spec: cloudrun.ServiceTemplateSpecArgs{
 					Containers: cloudrun.ServiceTemplateSpecContainerArray{
 						cloudrun.ServiceTemplateSpecContainerArgs{
-							Image: image.RepoDigest,
+							Image: image.RepoDigest.Elem(),
 							Resources: cloudrun.ServiceTemplateSpecContainerResourcesArgs{
 								Limits: pulumi.ToStringMap(map[string]string{
 									"memory": memory,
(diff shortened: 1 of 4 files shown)


container-gcp-go/main.go
```diff
@@ -97,7 +97,7 @@ func main() {
 				Spec: cloudrun.ServiceTemplateSpecArgs{
 					Containers: cloudrun.ServiceTemplateSpecContainerArray{
 						cloudrun.ServiceTemplateSpecContainerArgs{
-							Image: image.ImageName,
+							Image: image.RepoDigest,
```
cnunciato, line 100: When I run this, I get:

```
Diagnostics:
  pulumi:pulumi:Stack (cloudrun-test-dev):
    error: error in compiling Go: unable to run `go build`: exit status 1

    # cloudrun-test
    ./main.go:100:15: cannot use image.RepoDigest (variable of type pulumi.StringPtrOutput) as pulumi.StringInput value in struct literal: pulumi.StringPtrOutput does not implement pulumi.StringInput (missing method ToStringOutput)
```

I believe it should be:

```suggestion
							Image: image.RepoDigest.Elem(),
```

Look okay?

cnunciato (changes requested):
Just needs one fix I think, otherwise 👍.

container-gcp-go/main.go
cnunciato: Went ahead and applied this so we could get it merged. Thank you!

PR #702, opened 2023-12-06
[ci] enforce min CLI version of 3.95.0

dschaller:
mirrors https://github.com/pulumi/templates/pull/700 but for the release actions that are run only on merge


cnunciato (changes requested):
Sounds like we don't want, but let's discuss so we can get you what's needed here.

djgrove (commented):
Going to defer to @cnunciato on this one, "reviewing" to pop it off my PR queue ;)

cnunciato:
Closing this just to tidy up a bit, but feel free to reopen if you need.

PR #711, opened 2023-12-08
Set Go versions to 1.20

cnunciato:
This is the [minimum version we currently support](https://www.pulumi.com/docs/languages-sdks/go/), so this change installs Go 1.20 in CI and applies it to all `go.mod`s.


scottslowe (approved):
Approved, but worth noting that Go 1.22 is expected in Feb (at which point Go 1.20 will fall out of support).

PR #727, opened 2024-01-13
feat: add Pinecone serverless index templates

dirien:
Hi,

This PR adds starter templates for the Pinecone provider only for serverless indexes


pinecone-serverless-python/__main__.py
```diff
@@ -0,0 +1,16 @@
+"""An OpenStack Python Pulumi program"""
```
scottslowe, line 1:
```suggestion
"""A minimal Pinecone Serverless Python Pulumi program"""
```

scottslowe (changes requested):
This looks good (only one minor change needed on the Python program), but we'll need to include a C# template here as well (or follow up very quickly with a C# template).

dirien:
Hi @scottslowe,

thanks for the feedback and review! I must forgot the C# example, but now it is added! 

Looking forward for you review.

scottslowe (approved):

cnunciato:
@dirien Can we build a Java template also?

dirien:
> @dirien Can we build a Java template also?

@cnunciato we did not generated a Java SDK for the moment!

cnunciato:
How does one use these templates? When I run the TypeScript template with an API key for example, I get this:

```
Diagnostics:
  pulumi:pulumi:Stack (pinecone-serverless-typescript-03f0d18-dev):
    error: update failed

  pinecone:index:PineconeIndex (my-example-index):
    error: Failed to get Pinecone index: example-index-ts with http error code: 404
    error: failed to get Pinecone index: example-index-ts
```

Should we prompt for some of these values? Should we revisit the naming across these to be more consistent?

cnunciato:
Chatted with @dirien and the error may be because my account doesn't have early-access perms for this feature. #729 renames the templates `pinecone-$language`.

PR #741, opened 2024-02-01
fix: add LangServe AWS template

dirien:
This PR adds the LangServe AWS templates

Disclaimer: All other examples, get generated using `pulumi convert ...` Any changes should be marked in the `yaml` code


cnunciato (changes requested):
We currently support two kinds of templates in this repo -- starter templates and architecture templates. Starter templates are super-simple cloud-specific ways to bootstrap new projects, and architecture templates are cloud-specific ways to bootstrap common architectures. This is awesome, but it seems like neither of those; it's more like an example of how to deploy a specific application. 

pulumi/examples would therefore be a better place for this, I think. Programs can still be created (and even deployed with Deployments) as templates from that repo (by pointing to them by URL with `pulumi new`), so the user experience would still be the same -- you'd just call `pulumi new https://github.com/pulumi/examples/langserve-aws-ts` instead. Unless there's a super-strong overriding reason that this _needs_ to go into this repo, I'd suggest we house it there instead.

scottslowe:
@cnunciato These templates were specifically requested by @aaronkao, although I agree with your sentiment currently that these templates seem to fall somewhere between starter templates and architecture templates. I suppose it could be said that these are architecture templates (to deploy an AI app on LangServe), but I think we'd need to get more focused on the UX for these templates---with the other architecture templates, we had a strong "works out of the box with `pulumi up`" focus.

cnunciato:
@dirien Were you still planning on moving this to pulumi/examples?

PR #749, opened 2024-02-09
fix titles for installing the pulumi CLI

tgummerer:
The titles currently mention "Latest" pulumi CLI, however that's easily confused as the latest stable release, even though we are installing the dev versions now.

/cc @justinvp who noticed this


PollRobots (approved):
LGTM

PR #756, opened 2024-02-21
Use pulumi dev SDKs in CI tests

tgummerer:
Similarly how we're using the dev version of the pulumi CLI, we also want to use the dev version of the SDKs in the examples tests.  This will help us dogfood these better, and hopefully catch bugs before we release them.  We expect the dev versions to be stable and not introduce many new failures in CI here that otherwise wouldn't have been introduced.

To do this we're using the new `InstallDevReleases` flag, introduced in https://github.com/pulumi/pulumi/pull/15387

This is similar to https://github.com/pulumi/examples/pull/1595


PollRobots (approved):
LGTM

cnunciato (approved):

PR #762, opened 2024-03-01
specify pulumi version correctly

tgummerer:
This should fix the error seen in https://github.com/pulumi/templates/actions/runs/8116393836


cnunciato (approved):
LGTM, let's give it a try.

PR #765, opened 2024-03-04
free up disk space before tests

tgummerer:
These tests seem to occasionally run out of disk space.  Free up some space before running them to avoid that.


cnunciato (approved):
Thanks!

PR #771, opened 2024-03-25
ci: Use the latest dev version for cron runs

justinvp:
We're already using the dev version for PRs and performance cron runs, but aren't currently for this cron workflow. Which means I generally have to look at recent PRs to see how dev builds are doing before publishing a new pulumi/pulumi release. This PR updates the cron workflow to use the dev release.


tgummerer (approved):
Oh I missed that.  Thanks for updating it.

PR #774, opened 2024-04-12
Update TypeScript templates to include TypeScript 5 and target ES2020

julienp:
## Bump target to ES2020

Our minimum supported version of node and TypeScript both fully support ES2020 (https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-8.html#es2020-for-target-and-module, https://github.com/DefinitelyTyped/DefinitelyTyped/blob/0cebe5d5bcaad4a8f18e525b1cd098583b252585/types/node/v18/index.d.ts#L28C5-L28C31)

## Include TypeScript 5

We now fully support TypeScript 5 https://github.com/pulumi/pulumi/pull/15622. To make Pulumi use it instead of the bundled TS 3.8.3, it needs to be set as a dependency.

### DIFF
--- a/aiven-typescript/package.json
+++ b/aiven-typescript/package.json
@@ -5,7 +5,8 @@
         "@types/node": "^18"
     },
     "dependencies": {
-        "@pulumi/pulumi": "^3.20.0",
-        "@pulumi/aiven": "^5.4.0"
+        "@pulumi/aiven": "^5.4.0",
+        "@pulumi/pulumi": "^3.113.0",
+        "typescript": "^5.0.0"
     }
 }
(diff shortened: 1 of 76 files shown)


julienp:
Looks like the linux test failures are unrelated and match those for the previous PRs. I also did some spot checking with `pulumi new <url>` and it's all working as expected.

justinvp (approved):
LGTM

I called out places that could switch to spaces instead of tabs (while we're making changes here). But no need to block on that. We can always address those subsequently, if we wanted.

julienp:
> I called out places that could switch to spaces instead of tabs (while we're making changes here). But no need to block on that. We can always address those subsequently, if we wanted.

Definitely! I ran into a couple hat had mixed spaces and tabs and fixed those, but didn't notice that there were some with tabs only.

I'll do this in a separate PR so I can unblock the blog release before it gets late for me.

julienp:
https://github.com/pulumi/templates/pull/776

PR #789, opened 2024-05-10
Updates templates to use the Docker Build provider

desteves:


desteves:
I'm re-running the failed checks with debugging enabled, as the current log does not narrow down the issue. The only notable item that seems to impact multiple tests is along the lines of `environments.go:44: STDERR: Logging in using access token from PULUMI_ACCESS_TOKEN`

PR #797, opened 2024-06-04
Use secrets in Azure templates

TechWatching:


TechWatching (commented):
This will prevent people from having a secret displayed in their CI/CD logs when trying this template, while waiting for functions to properly tag outputs as secrets (cf. https://github.com/pulumi/pulumi/issues/12710)

flostadler:
This is still relevant, even after https://github.com/pulumi/pulumi/issues/12710 was closed because there's still some more work that's required to automatically mark those outputs as secret: https://github.com/pulumi/pulumi-azure-native/issues/1693.

@TechWatching I'll try to get this verified and merged now

flostadler:
the command dispatch seems to not be working correctly.

Triggered a test run manually: https://github.com/pulumi/templates/actions/runs/15437634385

Issue #814, opened 2024-07-22
update templates to use docker-build


mjeffryes:
docker-build is build is now the recommended way to build docker images, so we need to update our templates to use it.

mikhailshilkov:
It looks like this was supposed to be closed

PR #820, opened 2024-08-06
Update container templates to use docker-build provider

blampe:
We'd like to update templates to use the newer docker-build provider since it's recommended for building images going forward.

Fixes https://github.com/pulumi/templates/issues/814.


mjeffryes (approved):
Was surprised by the lack of an AWS version, but it looks like the use of docker is hidden by AWSX lib there.

PR #823, opened 2024-08-19
Update vm-gcp-python to used dictionary literals

julienp:
With type checking for dictionary literals, we now prefer this style over args classes https://github.com/pulumi/pulumi/issues/12689

### DIFF
--- a/.github/workflows/test-templates.yml
+++ b/.github/workflows/test-templates.yml
@@ -67,6 +67,12 @@ jobs:
       id-token: write
 
     steps:
+      - name: Free Disk Space (Ubuntu)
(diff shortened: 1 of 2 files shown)


.github/workflows/test-templates.yml
```diff
@@ -67,6 +67,12 @@ jobs:
       id-token: write
 
     steps:
+      - name: Free Disk Space (Ubuntu)
```
julienp, line 70: The tests seem to frequently run out of disk space. We use this actions in many of our repos to help with this, and it seems to work nicely here.

lunaris (approved):

PR #827, opened 2024-08-27
Update more python templates to use dictionary literals

julienp:
With type checking for dictionary literals, we now prefer this style over args classes https://github.com/pulumi/pulumi/issues/12689

### DIFF
--- a/kubernetes-python/__main__.py
+++ b/kubernetes-python/__main__.py
@@ -1,21 +1,19 @@
 """A Kubernetes Python Pulumi program"""
 
 import pulumi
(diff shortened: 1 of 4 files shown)


kubernetes-python/__main__.py
```diff
@@ -9,13 +9,13 @@
 
 deployment = Deployment(
     "nginx",
-    spec=DeploymentSpecArgs(
```
justinvp, line 12: Can we clean up some of the imports above?

justinvp (approved):
LGTM otherwise

PR #835, opened 2024-09-20
Update templates to use BucketV2

t0yv0:
Adopting our own best-practice recommendation per https://github.com/pulumi/home/issues/3632


t0yv0:
Reminder to self: before merging verify that Update static-website-* is covered by tests or test manually.

mjeffryes (approved):
TIL we have an f# aws template

t0yv0:
Some test failures are pertinent. I'll iterate here a bit.

t0yv0:
Still a bit more work here, will finish out on Monday.

PR #839, opened 2024-10-02
Update Azure Classic to v6

danielrbradley:
- Bump major version in dependencies.
- Add required subscriptionId config

Part of https://github.com/pulumi/pulumi-azure/issues/2336


thomas11 (commented):
What about azure-classic-visualbasic and azure-classic-yaml? The latter needs subscriptionId, too, doesn't it?

danielrbradley:
> What about azure-classic-visualbasic and azure-classic-yaml? The latter needs subscriptionId, too, doesn't it?

Good spot .. pushed. Not sure why my search didn't show them up.

thomas11 (approved):

PR #843, opened 2024-10-15
fix(github): correct export names

### LINKED ISSUE #842: Export name not suitable for github-{yaml,typescript} examples
The exports for the `github-yaml` and `github-typescript` examples are `clusterName`. This is presumably a copy & paste error from another template.

stooj:
Fixes #842


stooj:
This was solved in https://github.com/pulumi/templates/pull/846

PR #846, opened 2024-10-16
Converge all language examples for Github template to use the same output name.

thoward:
Fixes [#842](https://github.com/pulumi/templates/issues/842)


iwahbe (approved):
Thanks for cleaning this up here.

sean1588 (approved):

Issue #690, opened 2023-11-18
Update templates to target .NET 8 once it's been more widely adopted


justinvp:
Our .NET templates currently target .NET 6 (the previous only LTS version). Earlier this week, .NET 8 was released, the new LTS version. We should consider upgrading our templates to target .NET 8, so new projects have access to all the new .NET 8 features by default, once .NET 8 has been more widely adopted.

According to https://dotnet.microsoft.com/en-us/download/dotnet:
- .NET 6 will no longer be supported on November 12, 2024
- .NET 7 (which is not an LTS version) will no longer be supported May 14, 2024

Perhaps around May 14, 2024 would be a good time to move from targeting .NET 6 to .NET 8. We will definitely want to do it ahead of .NET 6 no longer being supported November 12, 2024.

pklaudat:
up

TechWatching:
It seems this has still not been done and November is coming soon. @justinvp  any plan to make it happen very soon?

aureq:
@justinvp any update on this please?

PR #849, opened 2024-11-01
Update dotnet templates to net8.0

smithrobs:
Addresses #690


justinvp:
@smithrobs, are you able to clone this repo directly and push these changes in a branch, and then open a PR from that branch into master? That should let the tests run.

I'm not sure what's going on with the `/run-templates` command.

PR #851, opened 2024-11-08
Update .NET templates to net8.0

### LINKED ISSUE #690: Update templates to target .NET 8 once it's been more widely adopted

justinvp:
This PR replaces #849

---

Support for .NET 6 is ending on November 12, 2024. This change updates templates to target .NET 8.

Fixes #690

### DIFF
--- a/.github/workflows/test-templates.yml
+++ b/.github/workflows/test-templates.yml
@@ -54,7 +54,7 @@ jobs:
         python-version:
           - 3.8
         dotnet-version:
-          - 6.0.x
+          - 8.0.x
         java-version:
           - 11
         java-distribution:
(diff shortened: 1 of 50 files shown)


justinvp:
The changes LGTM and the tests pass. Just need someone else to approve in order to merge.

julienp (approved):

interurban (approved):

PR #863, opened 2024-12-17
Add unit tests to aws-typescript

justinvp:
This change updates the `aws-typescript` template to include unit tests using jest. (Once we're happy with this, I'll add the moral equivalent for `aws-javascript` in a separate PR).

### DIFF
--- a/aws-typescript/jest.config.js
+++ b/aws-typescript/jest.config.js
@@ -1,0 +1,7 @@
+/** @type {import('ts-jest').JestConfigWithTsJest} **/
+module.exports = {
+    testEnvironment: "node",
+    transform: {
+        "^.+.tsx?$": ["ts-jest", {}],
+    },
+};
--- a/aws-typescript/tests/infra.spec.ts
+++ b/aws-typescript/tests/infra.spec.ts
@@ -1,0 +1,51 @@
+import * as pulumi from "@pulumi/pulumi";
+import "jest";
+
+// Test helper to convert a Pulumi Output to a Promise.
+// This should only be used in tests.
+function promiseOf<T>(output: pulumi.Output<T>): Promise<T> {
+    return new Promise(resolve => output.apply(resolve));
+}
+
+describe("infrastructure", () => {
+    // Define the infra variable as a type whose shape matches the that of the
+    // to-be-defined infra module.
+    let infra: typeof import("../infra");
+
+    beforeAll(() => {
+        // Put Pulumi in unit-test mode, mocking all calls to cloud-provider APIs.
+        pulumi.runtime.setMocks({
+            // Mock calls to create new resources and return a canned response.
+            newResource: (args: pulumi.runtime.MockResourceArgs) => {
+                // Here, we're returning a same-shaped object for all resource types.
+                // We could, however, use the arguments passed into this function to
+                // customize the mocked-out properties of a particular resource.
+                // See the unit-testing docs for details:
+                // https://www.pulumi.com/docs/iac/concepts/testing/unit/
+                return {
+                    id: `${args.name}-id`,
+                    state: args.inputs,
+                };
+            },
+
+            // Mock function calls and return an empty response.
+            call: (args: pulumi.runtime.MockCallArgs) => {
+                return {};
+            },
+        });
+    });
+
+    beforeEach(async () => {
+        // Dynamically import the infra module.
+        infra = await import("../infra");
+    });
+
+    // Example test. To run, uncomment and run `npm test`.
+    // describe("bucket", () => {
+    //     it("must have a name tag", async () => {
+    //         const tags = await promiseOf(infra.bucket.tags);
+    //         expect(tags).toBeDefined();
+    //         expect(tags).toHaveProperty("Name");
+    //     });
+    // });
+});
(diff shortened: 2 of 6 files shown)


aws-typescript/jest.config.js
justinvp, line 1: This is the standard jest config, configured to support ts-jest.

Note that this file could be `jest.config.ts` (TypeScript file), but that requires an explicit dependency on `ts-node`, which we don't currently have for this template (the vendored copy of `ts-node` in `@pulumi/pulumi` is used when running the pulumi program), so I've opted to leave it as a JavaScript file. Let me know if you think we should make it a TypeScript file for the TypeScript templates.

julienp (approved):
I noticed that the line numbers are wrong on test failures:

Screenshot 2024-12-18 at 17 03 19

I /think/ this is because we include the package `source-map-support` in pulumi, and that might cause the issue https://github.com/jestjs/jest/issues/10330, but I didn't look further than that.

aws-typescript/tests/infra.spec.ts
```diff
@@ -0,0 +1,51 @@
+import * as pulumi from "@pulumi/pulumi";
+import "jest";
+
+// Test helper to convert a Pulumi Output to a Promise.
+// This should only be used in tests.
+function promiseOf<T>(output: pulumi.Output<T>): Promise<T> {
+    return new Promise(resolve => output.apply(resolve));
```
julienp, line 7: I experimented with custom jest matchers here https://github.com/pulumi/templates/pull/868

This keeps the "magic" of Outputs, and avoids the (slightly wrong) conversion to promises here.

justinvp:
> I noticed that the line numbers are wrong on test failures:
> I /think/ this is because we include the package `source-map-support` in pulumi, and that might cause the issue [jestjs/jest#10330](https://github.com/jestjs/jest/issues/10330), but I didn't look further than that.

Nice catch! I tried uncommenting [our use](https://github.com/pulumi/pulumi/blob/efb38baee4dbdd41748df877dfa2daf0dfc11846/sdk/nodejs/index.ts#L16) of `source-map-support` in `@pulumi/pulumi` and that indeed fixed the line problem with jest. I wonder if we should remove our use of `source-map-support` and run node with the [`--enable-source-maps`](https://nodejs.org/docs/latest/api/cli.html#--enable-source-maps) option, since the capability is now built-in to Node since 12.12.0 per that issue.

julienp:
> Nice catch! I tried uncommenting [our use](https://github.com/pulumi/pulumi/blob/efb38baee4dbdd41748df877dfa2daf0dfc11846/sdk/nodejs/index.ts#L16) of `source-map-support` in `@pulumi/pulumi` and that indeed fixed the line problem with jest. I wonder if we should remove our use of `source-map-support` and run node with the [`--enable-source-maps`](https://nodejs.org/docs/latest/api/cli.html#--enable-source-maps) option, since the capability is now built-in to Node since 12.12.0 per that issue.

Nice, yeah we should absolutely do that! The version of ts-node we include in our *dev* dependencies also `includes source-map-support`, but that will not be included in the built package, and in any case I believe the key is removing the line you highlighted, which prevents source-map-support from activating when running under jest.

When running via the CLI, we'll usually run through `ts-node` (the vendored version, or the user installed version), which will enable source map support on its own. I *think* everything should be fine if we just remove *our* activation of it in `sdk/nodejs/index.ts`.

aws-typescript/tests/infra.spec.ts
julienp: I am not arguing for including these custom matchers in this PR. I was curious to see what it would look like. I think it's something we can follow up with if there are is enough demand to warrant the extra work. We could package the matchers in a library, that can be (optionally) used for writing tests.

aws-typescript/tests/infra.spec.ts
```diff
@@ -0,0 +1,51 @@
+import * as pulumi from "@pulumi/pulumi";
+import "jest";
+
+// Test helper to convert a Pulumi Output to a Promise.
+// This should only be used in tests.
```
julienp, line 5: I think we could elaborate a bit more here why this should only be used in tests, maybe something along the lines of:

```suggestion
// This should only be used in tests. Outputs track more information
// than Promises, such as the dependencies between resources.
// https://www.pulumi.com/docs/iac/concepts/inputs-outputs/#outputs
```

aws-typescript/tests/infra.spec.ts
```diff
@@ -0,0 +1,51 @@
...
+
+            // Mock function calls and return an empty response.
+            call: (args: pulumi.runtime.MockCallArgs) => {
+                return {};
+            },
+        });
+    });
+
+    beforeEach(async () => {
+        // Dynamically import the infra module.
```
julienp, line 39: We could mention that this needs to happen after the mocks have been setup, although it is already captured by the use of `beforeEach` vs `beforeAll`.

```suggestion
        // Dynamically import the infra module. This needs to happen
        // after the mocks have been setup.
```

tgummerer:
> Nice, yeah we should absolutely do that! The version of ts-node we include in our _dev_ dependencies also includes `source-map-support`, but that will not be included in the built package, and in any case I believe the key is removing the line you highlighted, which prevents source-map-support from activating when just importing `@pulumi/pulumi`, such as when running under jest.

Would this also fix https://github.com/pulumi/pulumi/issues/9218?

julienp:
> [pulumi/pulumi#9218](https://github.com/pulumi/pulumi/issues/9218)

Yes! I'll create a PR for this.

PR #864, opened 2024-12-17
Add unit tests to aws-python

justinvp:
This change updates the `aws-python` template to include unit tests, similar to https://github.com/pulumi/templates/pull/863 for `aws-typescript`.


aws-python/__main__.py
```diff
@@ -1,10 +1,6 @@
-"""An AWS Python Pulumi program"""
-
 import pulumi
-from pulumi_aws import s3
 
-# Create an AWS resource (S3 Bucket)
-bucket = s3.BucketV2('my-bucket')
+import infra
```
justinvp, line 3: Moves resource creation to a separate module, which is consistent with the change for `aws-typescript`. See https://github.com/pulumi/templates/pull/863#discussion_r1889302372.

aws-python/tests/test_infra.py
```diff
@@ -0,0 +1,42 @@
+import asyncio
+from typing import Awaitable, TypeVar
+
+import pulumi
+
+
+# Test helper to convert a Pulumi Output to an Awaitable.
+# This should only be used in tests.
+T = TypeVar("T")
+def awaitable(output: pulumi.Output[T]) -> Awaitable[T]:
```
julienp, line 10: Maybe `to_awaitable` ? `await awaitable(...)` reads a bit clumsy, and I find it a bit too close to uppercase `Awaitable`. I don't feel very strongly about this though.

Issue #865, opened 2024-12-17
[Epic] Add unit tests to project templates


justinvp:
## Overview

Augment project templates with unit tests, using our existing APIs for unit testing.

### Basic

- [ ] typescript
- [ ] javascript
- [ ] python
- [ ] csharp
- [ ] go
- [ ] java

### AWS

- [ ] https://github.com/pulumi/templates/pull/863
- [ ] aws-javascript
- [ ] https://github.com/pulumi/templates/pull/864
- [ ] https://github.com/pulumi/templates/pull/867
- [ ] https://github.com/pulumi/templates/pull/866
- [ ] aws-java

### Azure

- [ ] azure-typescript
- [ ] azure-javascript
- [ ] azure-python
- [ ] azure-csharp
- [ ] azure-go
- [ ] azure-java

### Google Cloud

- [ ] gcp-typescript
- [ ] gcp-javascript
- [ ] gcp-python
- [ ] gcp-csharp
- [ ] gcp-go
- [ ] gcp-java

### Kubernetes

- [ ] kubernetes-typescript
- [ ] kubernetes-javascript
- [ ] kubernetes-python
- [ ] kubernetes-csharp
- [ ] kubernetes-go
- [ ] kubernetes-java

PR #867, opened 2024-12-17
Add unit tests to aws-csharp

justinvp:
This change updates the `aws-go` template to include unit tests, similar to https://github.com/pulumi/templates/pull/863 for `aws-typescript`, https://github.com/pulumi/templates/pull/864 for `aws-python`, and https://github.com/pulumi/templates/pull/866 for `aws-go`.

Part of #865


aws-csharp/${PROJECT}.csproj
```diff
@@ -4,6 +4,7 @@
     <OutputType>Exe</OutputType>
     <TargetFramework>net8.0</TargetFramework>
     <Nullable>enable</Nullable>
+    <DefaultItemExcludes>$(DefaultItemExcludes);tests/**/*.cs;sdks/**/*.cs</DefaultItemExcludes>
```
justinvp, line 7: The more idiomatic thing to do would be to have the program and test projects SxS:

```
proj/
├── Pulumi.yaml
├── Program.cs
└── proj.csproj
tests/
├── InfraTests.cs
└── Tests.csproj
```

But we can't easily do that from a template. If we structured the template like this, you wouldn't be able to run `pulumi up` after `pulumi new` because the program would live in a subdirectory. You'd have to `cd` into the program's subdirectory first.

I think we want to keep it such that you can run `pulumi up` immediately after `pulumi new`, without having to change directories. So we nest the test project:

```
proj/
├── Pulumi.yaml
├── Program.cs
├── proj.csproj
└── tests/
    ├── InfraTests.cs
    └── Tests.csproj
```

To do so, we have to exclude the source files in the nested project here.

We have a similar issue with parameterized providers, where `pulumi package add` generates the sdks in a `sdks` subdirectory, and we ask users to add `<DefaultItemExcludes>` for that. For that reason, I've included excludes for both `sdks/` and `tests/` by default.

PR #868, opened 2024-12-18
[WIP] custom jest output matchers

julienp:
This shows a custom jest test matchers that know about Outputs.

`expect(...).apply(applyFn)` takes a callback that's meant to work like `Output.apply`. Internally it translates to a promise for jest so we don't have to deal with `done()`.

`expect(...).toEqualOutputOf(...)` can be used to easily check the resolved value of an output.

We could ship these matchers in a `@pulumi/jest-matchers` library.

### DIFF
--- a/aws-typescript/tests/infra.spec.ts
+++ b/aws-typescript/tests/infra.spec.ts
@@ -41,11 +41,20 @@ describe("infrastructure", () => {
     });
 
     // Example test. To run, uncomment and run `npm test`.
-    // describe("bucket", () => {
-    //     it("must have a name tag", async () => {
-    //         const tags = await promiseOf(infra.bucket.tags);
-    //         expect(tags).toBeDefined();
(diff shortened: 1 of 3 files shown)


aws-typescript/tests/infra.spec.ts
```diff
@@ -41,11 +41,20 @@ describe("infrastructure", () => {
...
-    //     });
-    // });
+    describe("bucket", () => {
+        it("must have a name tag", async () => {
+            const tags = await promiseOf(infra.bucket.tags);
+            expect(tags).toBeDefined();
+            expect(tags).toHaveProperty("Name");
+        });
+        it("must have the right tags", async () => {
+            await (expect(infra.bucket.tags) as any).toEqualOutputOf({ "Name": "My bucket oops" })
```
julienp, line 51: I didn't extend the `expect` type to know about our matchers, with that we won't need the cast to `any` here.

PR #891, opened 2025-03-09
Update scaleway templates to use Pulumiverse organization

Asamsig:


ringods (approved):

iwahbe:
In light of https://github.com/pulumi/templates/pull/947, we don't need the update anymore.

PR #904, opened 2025-04-29
Fix broken container-gcp-typescript program

jkodroff:
The container template for GCP in TS broke because of breaking changes upstream in the Docker Build provider.


jkodroff:
Verified that the template now works via:

```bash
pulumi new https://github.com/jkodroff/templates/tree/patch-1/container-gcp-typescript
```

justinvp (approved):

PR #923, opened 2025-06-04
Use secrets in Azure templates

flostadler:
Verification for https://github.com/pulumi/templates/pull/797


iwahbe:
Since #797 merged, I think we can close this.

PR #942, opened 2025-07-30
[hello-aws-javascript] Update index.js link to @pulumi/aws docs

pose:
Removing a 404 page from the template.


hello-aws-javascript/index.js
```diff
@@ -1,4 +1,4 @@
-// Import the [pulumi/aws](https://pulumi.io/reference/pkg/nodejs/@pulumi/aws/index.html) package
+// Import the [pulumi/aws](https://pulumi.com/reference/pkg/nodejs/@pulumi/aws/) package
```
tgummerer, line 1:
```suggestion
// Import the [pulumi/aws](https://www.pulumi.com/registry/packages/aws/api-docs/) package
```

I'm getting a 404 on the url you proposed as well.  I *think* the one here is the right one.

tgummerer (approved):

PR #943, opened 2025-07-30
Updating NodeJS to 22 (LTS version)

pose:
The [current](https://nodejs.org/en/about/previous-releases#looking-for-the-latest-release-of-a-version-branch) LTS version for NodeJS is 22. Updating to that version.

### DIFF
--- a/.github/workflows/test-templates.yml
+++ b/.github/workflows/test-templates.yml
@@ -50,7 +50,7 @@ jobs:
         go-version:
           - 1.21.x
         node-version:
-          - 18.x
+          - 22.x
         python-version:
           - 3.9
         dotnet-version:


.github/workflows/test-templates.yml
```diff
@@ -50,7 +50,7 @@ jobs:
         go-version:
           - 1.21.x
         node-version:
-          - 18.x
+          - 22.x
```
tgummerer, line 53: We should probably update to 20.x version here.  We support any node versions that are not EOL yet.  It's probably best to use the oldest supported version here, as it's more likely we break that, than a newer version.

pose: Thanks. Would you mind if I add to the test matrix `20` and `22`?

tgummerer: Having both sounds fine to me.

Frassle: 22 is LTS but we generally only test against oldest supported for these tests

pose: I added both and a link to where to find NodeJS release schedule.

pose: Updated to only use NodeJS 20.

julienp (approved):

PR #963, opened 2025-10-02
fix go.mod generator for azure-native and run it

tgummerer:
The azure-native SDK is split into multpile modules. When we updated the go.mod generator in https://github.com/pulumi/templates/pull/961, we missed that fact, and broke the azure-native templates go.mod. Fix that.

Fixes https://github.com/pulumi/pulumi-docker-containers/issues/549

### DIFF
--- a/azure-go/go.mod
+++ b/azure-go/go.mod
@@ -3,6 +3,7 @@ module ${PROJECT}
 go 1.23
 
 require (
-	github.com/pulumi/pulumi-azure-native/sdk/v3 v3.8.0
-	github.com/pulumi/pulumi/sdk/v3 v3.198.0
+	github.com/pulumi/pulumi-azure-native-sdk/resources/v3 v3.8.0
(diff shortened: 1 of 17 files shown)


azure-go/go.mod
```diff
@@ -3,6 +3,7 @@ module ${PROJECT}
 go 1.23
 
 require (
-	github.com/pulumi/pulumi-azure-native/sdk/v3 v3.8.0
-	github.com/pulumi/pulumi/sdk/v3 v3.198.0
+	github.com/pulumi/pulumi-azure-native-sdk/resources/v3 v3.8.0
+	github.com/pulumi/pulumi-azure-native-sdk/storage/v3 v3.8.0
+	github.com/pulumi/pulumi/sdk/v3 v3.199.0
```
tgummerer, line 8: This is the key change here, that restores what we had in before https://github.com/pulumi/templates/pull/961/files#diff-31c6663b34f8beca4dd7b7fdb13f45a96b6417b4ee9fad3e07792e217f9fd10cL6-L8, except with a higher version number.

PR #965, opened 2025-10-03
Remove toolchain directives from go templates

justinvp:
These shouldn't be necessary for templates. Also, update CI to use the oldest supported version of go (1.24.x).


iwahbe (approved):
Changes look good. 

CI failures look like https://github.com/pulumi/pulumi-service/issues/33237, rather then a problem introduced here.

PR #973, opened 2025-10-16
Specify pip/venv for GCP

joeduffy:


justinvp (changes requested):
As discussed in Slack, I don't think we should take this change.

PR #993, opened 2025-11-19
Delete `metadata` and supporting files

iwahbe:
These files lead to toil (for example: https://github.com/pulumi/templates/pull/968), but they don't actually contribute to the repository AFAIK.

They were added to facilitate https://github.com/pulumi/pulumi-service/pull/10486, but the service doesn't use them as of https://github.com/pulumi/pulumi-service/issues/34338.


iwahbe:
Looks like we still use this in https://github.com/pulumi/pulumi-service/blob/4c85ee3065790c374a99b234c9300bdedefedaf1/cmd/service/api/organizations_templates.go#L594-L595, so we need to keep this for now.

PR #1000, opened 2025-11-25
Add Kotlin templates for AWS, GCP and Azure

morhaf-ko:
Adding kotlin templates


iwahbe:
Sorry for the very long wait for a response. We are not accepting new templates right now.

PR #1141, opened 2026-06-23
Add HCL versions of all architecture and cloud starter templates

cnunciato:
Adds 22 new HCL templates that use our `pulumi/*` providers:

- Starter templates: `hcl`, `aws-hcl`, `azure-hcl`, `gcp-hcl`, `kubernetes-hcl`
- Architecture templates: Static website, serverless, VM, container Kubernetes cluster, and Kubernetes app, across all clouds

Accompanying PR to add them to pulumi.com/templates: https://github.com/pulumi/docs/pull/19831

## Validation

Each template was exercised with a full `pulumi up` → `destroy` on a real cloud (AWS via ESC, Azure, GCP) or a local cluster (Kubernetes-application templates).

## Summary

* ✅ 19 templates fully working
* ⚠️ `serverless-aws-hcl` deploys, but its `url` output fails to resolve: when an `aws-apigateway:RestAPI` is created with an `event_handler` (Lambda) route, all of the component's outputs collapse to a placeholder, so `url` comes back empty. A RestAPI created with only static routes resolves `url` correctly — so it's a create-time output-resolution bug triggered by an `event_handler` route. Filed https://github.com/pulumi-labs/pulumi-hcl/issues/298 to track.
* ❌ `vm-azure-hcl` and `kubernetes-azure-hcl` fail on `pulumi preview`, crashing with `fatal error: stack overflow`. Filed https://github.com/pulumi-labs/pulumi-hcl/issues/290 to track.

Part of https://github.com/pulumi/marketing/issues/1771.


iwahbe:
Pulumi HCL can consume Azure Native directly:

```hcl
terraform {
  required_providers {
    azure = {
      source  = "pulumi/azure-native"
    }
  }
}
```

I don't see a need to stick to only providers that are in the opentofu registry. We can do the same for docker-build & kubernetes.

iwahbe:
This PR will need to rebase on top of https://github.com/pulumi/templates/pull/1143 to pass CI.

cnunciato:
> I don't see a need to stick to only providers that are in the opentofu registry.

Opened up an internal discussion on it. My hunch is that keeping as close as we can to what Terraform and OpenTofu users will expect (at least for now) is probably the way to go with these, but open to talking through it. Keeping to the `hashicorp` et. al.  set will also help us fend off any incorrect first impressions that our HCL support _requires_ Pulumi providers, which of course it doesn't.

cnunciato:
Circling back: After some discussion, we decided the best default experience would be to use `pulumi/*` providers, and then document our support for `hashicorp/*`, etc., providers so it's clear we support them.

container-azure-hcl/app/Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM node:16
```
jkodroff: v24 is latest LTS

container-gcp-hcl/app/Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM node:16
```
jkodroff: v24 is current LTS.

jkodroff (approved):
Looks fine - I scanned rather than going over with a fine-toothed comb. Noted that the node version used is rather old at this point.

container-azure-hcl/app/Dockerfile
cnunciato: Good eye, thanks! Bumping the others as well.

PR #1175, opened 2026-08-17
[lockfiles] Commit a lockfile for every Node template

fnune:
## Why

This is a first line of defense for the customer who runs `pulumi new`, against a supply-chain attack on the packages a template pulls in.

Node templates declare open ranges like `"@pulumi/aws": "^7.0.0"` and ship no lockfile, so every `pulumi new` installs whatever is newest that day. Nobody can review that, because nothing here records what a user gets.

A lockfile pins the whole tree. Every user gets the same packages, and a version change becomes a reviewable diff.

Internal: [Improving `pulumi new` supply chain security](https://app.notion.com/p/Improving-pulumi-new-supply-chain-security-391fdbdf1cce80fabd18c38d6a7632bf)

## Commands

| Command | What it does |
| --- | --- |
| `make lockfiles` | Rebuilds every lockfile from its `package.json`. Run it after you edit a `package.json`. |
| `make audit_lockfiles` | Fails on a vulnerability this branch introduces. |
| `make test_lockfiles` | Checks that a published template still carries its lockfile. |

## Flow

```mermaid
flowchart TD
    R(["Renovate opens a pull request"]) --> CI
    E(["You edit a template's package.json"]) --> ML["You: make lockfiles, commit"]
    ML --> CI

    CI["CI: make audit_lockfiles,<br/>make lockfiles, make test_lockfiles"] --> Q{"All green?"}
    Q -->|"yes"| M["Merged"]
    Q -->|"no"| H["You: bump the dependency in package.json,<br/>or override the transitive one there"]
    H --> CI

    classDef bad fill:#fee2e2,stroke:#dc2626,color:#450a0a
    classDef good fill:#dcfce7,stroke:#16a34a,color:#052e16

    class H bad
    class M good
```

Renovate owns the refresh. `lockFileMaintenance` re-resolves the pins weekly to the newest version each declared range allows, which is where a published fix normally sits. The shared config waits three days before it adopts a release, and skips versions OSV reports as malicious.

CI never lets a branch make things worse. `make audit_lockfiles` compares against the base branch, so a vulnerability somebody else introduced never blocks you.

## Checks

1. **No vulnerability this branch introduces.** Audited against the base branch.
2. **The lockfile matches its `package.json`.** CI regenerates every lockfile and fails on any difference.
3. **The lockfile reaches the user.** `pulumi template publish` obeys `.gitignore`, so an ignored lockfile never leaves the repo and no diff shows it. `tests/template_publish_test.go` runs the real publish command against a backend that captures the archive. It reimplements nothing, so it cannot drift from the command.
4. **The template still deploys.** The acceptance suite deploys these for real.

## npm and bun

Both are first-class. Each project is audited by the tool that wrote its lockfile, because neither reads the other's. Renovate keeps a separate `bun` manager, so `renovate.json5` enables both.

## Not covered

**pnpm and yarn users of a `nodejs` template.** A template directory can carry one lockfile only. `ResolvePackageManager` picks the package manager by which lockfile it finds, so adding `pnpm-lock.yaml` hands every user of that template to pnpm. Those users need a separate template that declares `packagemanager`, which is a catalogue decision.

**Templates in other languages.** They ship no lockfile. Every one of the 141 open critical alerts on this repo comes from a Go template, none from a Node one, and `golang.org/x/crypto` accounts for 102 of them.


pose (changes requested):
Hi @fnune, thanks for putting this PR together. I am bit confused why we need to do so much manual work. Is there a way we could use renovate here instead of introducing new Makefiles targets and scripts? I was under the impression that Renovate could keep updating the lockfiles for us (and even generate them), so we don't need to do it manually. Also, the GH Actions seem to be as well something that Renovate could do for us. Thanks!

fnune:
I wasn't aware Renovate can do some of this. I'll take a look and say something tomorrow!

pose:
> I wasn't aware Renovate can do some of this. I'll take a look and say something tomorrow!

Yes, specially since we have good security defaults on what packages we are allowed to fetch and how fast security updates flow through the pipeline. Happy to spend some time going over it together as I don't consider myself an expert on Renovate.

iwahbe (commented):
I'd much prefer handling this via renovate if we need to make a change at all.

I'd like to see a reason to make a change like this more substantial then "to pin the tree".

fnune:
@iwahbe The reason is providing a first line of defense for customers instantiating templates against supply-chain attacks, by providing an audited lockfile.

Internal: https://app.notion.com/p/Improving-pulumi-new-supply-chain-security-391fdbdf1cce80fabd18c38d6a7632bf

pose (changes requested):
I really would like to get to the bottom on why Renovate cannot do this 🙏

fnune:
Renovate is doing its job here: keeping the lockfiles up to date periodically.
https://github.com/pulumi/templates/pull/1175/changes#diff-31b1aa2b93afe11792d84c6cb5057a064b5a3294fa0caab92d7570ea604f759f

Makefile
```diff
@@ -66,5 +66,17 @@ test_templates::
 metadata::
 	yarn && yarn run metadata && yarn test
 
+
+lockfiles::
+	./scripts/generate-lockfiles.sh
+
+
+test_lockfiles::
```
pose, line 78: test packaging/publishing includes those file

scripts/audit-lockfiles.sh
iwahbe, line 1: I don't think we need or want this. I'm assuming that renovate will only pull us to non-retraced versions and Dependabot should warn us if we have pinned a retracted version. We don't want or need to re-implement those systems here.

scripts/generate-lockfiles.sh
```diff
@@ -0,0 +1,125 @@
...
+        }
+        /^runtime:[[:space:]]*$/ { in_runtime_block = 1; next }
+        in_runtime_block && /^[^[:space:]]/ { exit }
+        in_runtime_block && /^[[:space:]]+name:[[:space:]]*/ {
+            sub(/^[[:space:]]+name:[[:space:]]*/, "")
+            sub(/[[:space:]]+$/, "")
+            print
+            exit
+        }
+    ')"
```
iwahbe: Let's write this script in Python, where we can just unmarshal the YAML.

tests/template_publish_test.go
```diff
@@ -0,0 +1,173 @@
...
+	sort.Strings(templates)
+
+	for _, template := range templates {
+		t.Run(template, func(t *testing.T) {
+			published := publishTemplate(t, filepath.Join(root, template), template)
+
+			for _, lockfile := range lockfiles[template] {
+				assert.Contains(t, published, lockfile,
+					"pulumi template publish drops %s/%s, most likely because a .gitignore entry excludes it, so it never reaches the users who install this template",
+					template, lockfile)
```
iwahbe, line 62: Why do we want to assert this? Don't we want to assert the opposite?

iwahbe (commented):
Does this change mean that pnpm & yarn users are going to see a package.lock and a warning on first use?

Does pulumi-language-nodejs handle converting out of package.lock correctly? If not, I think this is blocked until it does. If so, then your "Not Covered" comment is wrong.

scripts/generate-lockfiles.sh
fnune: Yeah was discussing this just now with Alberto

fnune:
Note from call: add a test to CI when whe run `pulumi new` to check that the CLI doesn't write over the lockfiles or ignore them.

.github/workflows/audit-lockfiles.yml
```diff
@@ -0,0 +1,31 @@
...
+        with:
+          node-version: 24.x
+
+      - name: Install Bun
+        uses: oven-sh/setup-bun@v2
+        with:
+          bun-version: 1.4.0
+
+      - name: Audit lockfiles for vulnerabilities this change introduces
+        run: ./scripts/audit-lockfiles.sh --base "${{ github.event.pull_request.base.sha }}"
```
pose: https://gitlab.com/tanna.dev/renovate-graph

pose (approved):

scripts/audit-lockfiles.sh
fnune: Agree on not wanting to re-implement things. I'm going to try and do this with Renovate as well.

The reason there is a separate script and CI step is that there is that Renovate is configured to look at the default branch only. I want to avoid us merging a pull request that contains a bump to a vulnerable or malicious version. To avoid that, I need a script that runs on PRs and blocks merging. Renovate can't do that as configured.

However, I'm going to try to figure out how to use Renovate on demand for PRs instead of having this script.

tests/template_publish_test.go
fnune: We're asserting that the lockfile is included in the published artifact. That's wanted. Maybe the error message string confused you? It shows only if the assertion fails.

fnune:
> Does this change mean that pnpm & yarn users are going to see a package.lock and a warning on first use?

Yes, we discussed this as a worthwhile trade-off of shipping vetted `package-lock.json` files:
https://app.notion.com/p/Improving-pulumi-new-supply-chain-security-391fdbdf1cce80fabd18c38d6a7632bf?source=copy_link#77bec2a457424421b4d54fd0868bdc5a

fnune:
> Does pulumi-language-nodejs handle converting out of package.lock correctly? If not, I think this is blocked until it does. If so, then your "Not Covered" comment is wrong.

Will investigate, I'm not sure. Good callout!

julienp:
> > Does pulumi-language-nodejs handle converting out of package.lock correctly? If not, I think this is blocked until it does. If so, then your "Not Covered" comment is wrong.
> 
> Will investigate, I'm not sure. Good callout!

It /might/ for bun, although it will leave the package-lock.json around, which we'd have to delete. I think for pnpm we would have to run `pnpm import`. Also not sure if this works for all versions of package-lock.json.. I don't think yarn classic can do anything with recent package-lock.json format.

pose (approved):

